Issue #9, opened 2019-05-20
[improvement] optimize images as much as possible


atb00ker:
The build time and more importantly the image size should be reduced as much as possible.
- [x] Reduce the number of layers.
- [x] Remove apk and python cache.
- [x] Find and remove all the build dependencies.
- [x] Leverage multistage build and reduce size further

**Research:**
- https://bit.ly/2GBtVtI
- https://github.com/psycopg/psycopg2/issues/684
- Python multistage benefits ([See here](https://blog.realkinetic.com/building-minimal-docker-containers-for-python-applications-37d0272c52f3))

atb00ker:
Closing for now because base image is optimized, will open again in face of new possibility of optimizations. :smile:

PR #23, opened 2019-05-21
[testing] Added basic travis-ci tests

atb00ker:
- Initial travis tests
- Changes in docker images to ensure that tests don't take too long. (A following PR will be dedicated to build-time improvements.)

Related: #9 #24 #29


.travis.yml
```diff
@@ -1,9 +1,24 @@
...
+jobs:
+  include:
+    - stage: tests
+      script:
+      # Start containers
+      - docker-compose build
+      - docker-compose up -d
+      # Getting ready to test
+      - echo "127.0.0.1 dashboard.openwisp.org controller.openwisp.org radius.openwisp.org topology.openwisp.org" | sudo tee -a /etc/hosts
+      - sleep 1m
```
nemesisdesign: why sleep for 1 minute? Is not clear. Maybe we need a comment, or maybe we don't need to wait 1 minute, or maybe there's a better way to understand when we can get started to test, @2stacks @hispanico do you have any suggestion on this front?

build/openwisp_radius/Dockerfile
```diff
@@ -1,15 +1,22 @@
 FROM python:3.7-alpine AS BASE
 
-WORKDIR /opt/openwisp/
-RUN apk add  --update --no-cache build-base postgresql-dev libffi-dev openssl-dev python3-dev zlib-dev jpeg-dev git gettext linux-headers
-RUN pip install https://github.com/openwisp/django-freeradius/tarball/master && \
+RUN apk add  --update --no-cache \
+    # TODO: purge the build-deps after python installation.
+    libffi-dev build-base postgresql-dev openssl-dev python3-dev zlib-dev jpeg-dev \
+    git gettext linux-headers && \
```
nemesisdesign: it seems many of these packages are defined in each image, I think we need a base image from which all images inherit from to avoid this repetition. Building may also be faster if we don't change the base image often.

@atb00ker what do you think? In some projects I work on, we almost always have a base image.

build/openwisp_dashboard/Dockerfile
```diff
@@ -1,13 +1,18 @@
 FROM python:3.7-alpine AS BASE
 
-WORKDIR /opt/openwisp/
-RUN apk add --update --no-cache build-base libffi-dev openssl-dev postgresql-dev python3-dev zlib-dev jpeg-dev musl-dev gettext linux-headers git --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing gdal-dev geos-dev
-
 COPY ./openwisp_dashboard/requirements.txt /opt/openwisp/requirements.txt
-RUN pip install -r requirements.txt
+RUN apk add --update --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing \
```
nemesisdesign: is it good to use `http://dl-cdn.alpinelinux.org/alpine/edge/testing`?

From https://wiki.alpinelinux.org/wiki/Edge

`Warning: "edge" is under constant development so be careful using it in production. It is possible that bugs in "edge" could cause data loss or could break your system.`

@2stacks @hispanico what do you think?

It seems to me that we should use something safer.

tests/tests.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+function dashboard_admin_login {
+    touch $COOKIES
+    $CURL_BIN $LOGIN_URL > /dev/null
+    DJANGO_TOKEN="csrfmiddlewaretoken=$(grep csrftoken $COOKIES | \
+    sed 's/^.*csrftoken[[:blank:]]*//')"
+    $CURL_BIN -d "$DJANGO_TOKEN;username=$USERNAME;password=$PASSWORD" \
+    -X POST $LOGIN_URL
```
nemesisdesign: increase the indentation to make sure is clear it's a line continuation, eg:

```
$CURL_BIN -d "$DJANGO_TOKEN;username=$USERNAME;password=$PASSWORD" \
          -X POST $LOGIN_URL
```

tests/tests.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+function dashboard_admin_login {
+    touch $COOKIES
+    $CURL_BIN $LOGIN_URL > /dev/null
+    DJANGO_TOKEN="csrfmiddlewaretoken=$(grep csrftoken $COOKIES | \
+    sed 's/^.*csrftoken[[:blank:]]*//')"
```
nemesisdesign: same here with the indentation

tests/tests.sh
nemesisdesign: this is great, but you should check the response in some way, as you're doing in the other tests, otherwise it looks like (at least to me) that it won't help much, what do you think?

tests/tests.sh
```diff
@@ -0,0 +1,33 @@
...
+    $CURL_BIN -d "$DJANGO_TOKEN;username=$USERNAME;password=$PASSWORD" \
+    -X POST $LOGIN_URL
+}
+
+function run_dashboard_tests {
+    # This tests the login was successful being able to 
+    # access this page also means redis is running.
+    $CURL_BIN --head http://dashboard.openwisp.org/admin/ | grep -q "200 OK" && \
+    echo "SUCCESS: Admin login successful!" || \
+    (echo "ERROR: Admin login failed!"; exit 1)
```
nemesisdesign: fix the indentation here

tests/tests.sh
```diff
@@ -0,0 +1,33 @@
...
+    sed 's/^.*csrftoken[[:blank:]]*//')"
+    $CURL_BIN -d "$DJANGO_TOKEN;username=$USERNAME;password=$PASSWORD" \
+    -X POST $LOGIN_URL
+}
+
+function run_dashboard_tests {
+    # This tests the login was successful being able to 
+    # access this page also means redis is running.
+    $CURL_BIN --head http://dashboard.openwisp.org/admin/ | grep -q "200 OK" && \
+    echo "SUCCESS: Admin login successful!" || \
```
nemesisdesign: great! But fix the indentation please

tests/tests.sh
```diff
@@ -0,0 +1,33 @@
...
+function init_dashoard_tests {
+    LOGIN_URL=http://dashboard.openwisp.org/admin/login/
+    USERNAME='admin'
+    PASSWORD='admin'
+    COOKIES=cookies.txt
+    CURL_BIN="curl -s -c $COOKIES -b $COOKIES -e $LOGIN_URL"
+    # Check if openwisp-dashboard is reachable
+    curl -s --head $LOGIN_URL  | grep -q "200 OK" && \
+    echo "SUCCESS: openwisp-dashboard login page reachable!" || \
+    (echo "ERROR: openwisp-dashboard login page not reachable!"; exit 1)
```
nemesisdesign: indentation

tests/tests.sh
```diff
@@ -0,0 +1,33 @@
...
+
+function init_dashoard_tests {
+    LOGIN_URL=http://dashboard.openwisp.org/admin/login/
+    USERNAME='admin'
+    PASSWORD='admin'
+    COOKIES=cookies.txt
+    CURL_BIN="curl -s -c $COOKIES -b $COOKIES -e $LOGIN_URL"
+    # Check if openwisp-dashboard is reachable
+    curl -s --head $LOGIN_URL  | grep -q "200 OK" && \
+    echo "SUCCESS: openwisp-dashboard login page reachable!" || \
```
nemesisdesign: indentation

tests/tests.sh
```diff
@@ -0,0 +1,33 @@
...
+    (echo "ERROR: openwisp-dashboard login page not reachable!"; exit 1)
+    dashboard_admin_login # Login to django-admin
+    run_dashboard_tests # Run CI tests
+    rm $COOKIES # Delete cookies file created during testing
+}
```
nemesisdesign: if you can write a test that creates a new user in the system, that would be SUPER, because we had many issues with that recently

build/openwisp_radius/Dockerfile
2stacks: I believe postgis dependencies for alpine are only available in edge.

tests/tests.sh
2stacks: Not sure if its what your looking for but I use the django-freeradius batch_add_user in my travis tests.
>docker-compose run --rm -v $PWD/scripts/users.csv:/users.csv django python manage.py batch_add_users --name users --file /users.csv

nemesisdesign: I meant to add an user from the web interface

.travis.yml
atb00ker: Indeed. Sleep is there to ensure all the services are running.
Not just the postgresql but also the `migrations` and `collectstatic`.
Another way would be loop a cURL request until I get 200 on the login page. :-)

build/openwisp_radius/Dockerfile
atb00ker: I am working on the base image right now. :-)

tests/tests.sh
atb00ker: That's in the kanban board, I will implement that after the postfix service container is ready and running! :-)

build/openwisp_dashboard/Dockerfile
atb00ker: Currently `gdal-dev geos-dev` are only available in `edge`, alternative would be to compile them, I was initially compiling them but that takes a lot of time so for convenience I moved to this. 
Let me know if compiling them sounds better. :-)

build/openwisp_base/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+# ENV PYTHONPATH=/install/lib/python3.7/site-packages
+# RUN apk add --update --virtual .build-deps postgresql-dev git
+# RUN pip install asgi_redis service_identity django-redis psycopg2 uwsgi openwisp-network-topology
+# RUN pip install https://github.com/openwisp/django-netjsonconfig/tarball/master
+# RUN pip install https://github.com/openwisp/django-x509/tarball/master
+# RUN pip install https://github.com/openwisp/django-freeradius/tarball/master
+# RUN pip install https://github.com/openwisp/openwisp-radius/tarball/master
+# RUN pip install https://github.com/openwisp/openwisp-users/tarball/master
+# FROM BASE
+# COPY --from=PIP_INSTALL /usr/local /usr/local
```
2stacks: My only suggestion is to include anything else that is common to all containers.  It looks like this may be a candidate
>COPY ./common/ /opt/openwisp/

docker-compose.yml
```diff
@@ -1,9 +1,26 @@
...
+    deploy:
+      replicas: 1
+    environment:
+      - DB_ENGINE=$DB_GIS_ENGINE
+    env_file:
+      - .env
+    volumes:
+      - openwisp_static:/opt/openwisp/static
+      - openwisp_media:/opt/openwisp/media
+
```
2stacks: Are the build/args, deploy and volumes syntax needed for the base container?  It looks as though its intended to be a running container vs a base for building other containers.

nemesisdesign: oh yeah, I also use `Makefile` a lot, it's a shame I did not start using this in OpenWISP.
If we can use a `Makefile` to initialize the build, let's do it, so we centralize the logic in one source of truth.

.travis.yml
nemesisdesign: the loop can work with a time out or max number of attempts, I like that better than the sleep

build/openwisp_controller/Dockerfile
```diff
@@ -1,21 +1,15 @@
-FROM python:3.7-alpine AS BASE
+FROM atb00ker/ready-to-run:openwisp-base
```
nemesisdesign: what is `atb00ker/ready-to-run`?

I would like this to be in the openwisp organization since the beginning. I think @hispanico created one a while ago.

build/openwisp_base/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM python:3.7-alpine AS BASE
+
+COPY ./openwisp_base/requirements.txt /opt/openwisp/requirements.txt
```
nemesisdesign: don't you need to also install these dependencies? I would also move this command after the `RUN apk add --update --no-cache \ ...etc...` because the lines in `requirements.txt` may change more often than the system packages, so if we need to add one python package at least the system packages won't have to be rebuilt again.

build/openwisp_base/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN apk add --update --no-cache \
+    --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing \
+    gdal-dev geos-dev zlib-dev jpeg-dev openssl-dev libffi-dev gettext gcc && \
+    apk add --update --virtual .build-deps postgresql-dev git build-base linux-headers && \
+    pip install --no-cache-dir -r /opt/openwisp/requirements.txt && \
```
nemesisdesign: I see now that you do everything in one step, are you doing it to save on image size?

I would separate the part which deal with system packages and the part which deals with the python packages

tests/tests.sh
nemesisdesign: great :+1: thanks for clarifying

docker-compose.yml
atb00ker: > Are the build/args, deploy and volumes syntax needed for the base container?

Oops, no they are not required at all.

build/openwisp_base/Dockerfile
atb00ker: It's done in one step because `.build-deps` are installed and uninstalled in the same layer to avoid taking useless files to next layer. It's a common practice I've noticed in a lot of other images as well. 
The system packages not in `.build-deps` don't take a lot of time to install, the real thing that takes a lot of time is `pip install` only, hence, I coupled all of them in one layer.

Should I keep `.build-deps` with python packages only or move them seperately as well?
(That makes dockerfiles look cleaner but bloats the image to about twice the size.)

build/openwisp_base/Dockerfile
atb00ker: Hi,

Are you taking about: `COPY --from=PIP_INSTALL /usr/local /usr/local` here?
I was testing it to achieve something like Multistage builds [here](https://blog.realkinetic.com/building-minimal-docker-containers-for-python-applications-37d0272c52f3).

(If you mean some other line than the highlighted one or I've misunderstood, please let me know.)

2stacks: No sorry I'm referring to this line in the other Dockerfiles https://github.com/openwisp/docker-openwisp/blob/646147986fb0652a7ca972395ba43d4686b4f03b/build/openwisp_controller/Dockerfile#L5

docker-compose.yml
atb00ker: Please review the Makefile. :-)

build/openwisp_base/Dockerfile
atb00ker: Please can you check again if I understood you correctly. :-)

2stacks: Yes I believe this takes care of it. https://github.com/openwisp/docker-openwisp/blob/fe4f0a3849e5c52b3a522f655e0e5bc822f27c29/build/openwisp_base/Dockerfile#L13

build/openwisp_base/Dockerfile
nemesisdesign: how much size increase are we talking about? If it's less than 100 MB, proceed to separate them.

build/openwisp_controller/Dockerfile
nemesisdesign: I created this: https://github.com/openwisp/docker-openwisp/issues/29
We can discuss it there and solve it next week.

build/openwisp_base/Dockerfile
atb00ker: After the new changes, the current size is ~450MBs and seperating will make it around ~600MBs.
(However, since this is the base image, all the derived image size will be increased.)

atb00ker:
@nemesisdesign Alrighted, Updated! :+1:

PR #37, opened 2019-06-04
[enchancement] Postfix image

### LINKED ISSUE #22: [tasklist] postfix image
> Provide a deafult working postfix instance which is able to send emails but can also be turned off (users may want to use an external SMTP service)

atb00ker:
Add postfix image

Remaining:
- [x] write tests to check if postfix is running properly.
- [ ] Unexpected behaviour while adding user -- works even without postfix

Closes #22


atb00ker:
Hello,

I am facing an issue with adding a new user:
Even when I turn off the postfix container, I am able to add a new user as if there were no errors.
I see an error while trying to send a password recovery mail due to absense of postfix but not while creating a new user. Any pointeres on getting that error would be help.

Thanks.

PR #39, opened 2019-06-12
[improvement] Configurability + remove redundancy

atb00ker:
- Add sentry for django.
- Add leftlet configuration options.
- Moved common environment variables to openwisp base image.
- Moved postfix "init_command" script with the init_command script of other 
services to have a common script for all. (just following the pattern of other services.)


build/common/openwisp/settings.py
```diff
@@ -181,6 +194,18 @@
...
 }
 
+# Sentry
+# https://sentry.io/for/django/
+
+if os.environ['DJANGO_SENTRY_DSN']:
+    import sentry_sdk
+    from sentry_sdk.integrations.django import DjangoIntegration
+    sentry_sdk.init(dsn=os.environ['DJANGO_SENTRY_DSN'],
+                    integrations=[DjangoIntegration()])
```
nemesisdesign: :+1:

build/common/openwisp/utils.py
```diff
@@ -11,6 +11,11 @@ def filter(self, record):
         record.host = socket.gethostname()
         return True
 
+def env_bool(env):
+    if env in ["True", "true", "TRUE"]:
```
nemesisdesign: what about:

```python
return env in ["True", "true", "TRUE"]
```

build/common/utils.sh
```diff
@@ -66,3 +68,83 @@ function envsubst_create_config {
         envsubst < ${1} > /etc/nginx/conf.d/${application}.${2}.conf
     done
 }
+
+function postfix_config {
```
nemesisdesign: is it possible to disable this container in case we use an external SMTP? Ensure it's explained in the README how to do this.

build/openwisp_websocket/daphne.conf
```diff
@@ -14,3 +14,5 @@ autorestart=true
 stopsignal=INT
 stdout_logfile=/dev/stdout
 stderr_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+stderr_logfile_maxbytes=0
```
nemesisdesign: why are these needed? If there's a reason, we may need to add a comment because it's not clear to me

build/openwisp_websocket/worker.conf
```diff
@@ -14,3 +14,5 @@ autorestart=true
 stopsignal=INT
 stdout_logfile=/dev/stdout
 stderr_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
```
nemesisdesign: same here

build/openwisp_websocket/daphne.conf
atb00ker: It's required because otherwise we get `error: invalid seek`, it's not a fatal error that stops the service but it ruins the logs.
I will add a comment here. :smile:

build/common/utils.sh
atb00ker: Added a note in the README. :-)

build/openwisp_websocket/daphne.conf
nemesisdesign: ok, let me know when is done

atb00ker: Added a commit above this line. :smile:

build/common/utils.sh
atb00ker: Added a section in README.

Issue #43, opened 2019-07-06
[bug] create_prod_certs not working


atb00ker:
The function to create production certificates have errors and is not working properly.

atb00ker:
Solved in #44

PR #44, opened 2019-07-12
 [enchancement] Add freeradius

### LINKED ISSUE #16: [tasklist] freeradius image
> Provide a configurable freeradius instance in a dedicated container

atb00ker:
Working on:
- [x] Kubernetes files
- [x] Add Testcases
- [x] Serving freeradius behind Nginx

Closes #16 #43


build/openwisp_freeradius/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+ENTRYPOINT ["sh", "init_command.sh"]
+EXPOSE 1812/udp 1813/udp
+
+COPY ./common/init_command.sh \
+    ./common/utils.sh ./
+COPY --chown=root:root ./openwisp_freeradius/raddb/ /etc/raddb/
+
+ENV MODULE_NAME=freeradius \
+    DASHBOARD_DOMAIN=dashboard.example.com \
```
2stacks: Does this mean the freeradius container will attempt to connect to the Rest API on the dashboard container?  Shouldn't it connect to the openwisp_radius container?

docker-compose.yml
```diff
@@ -134,6 +141,19 @@ services:
...
+    build:
+      context: build
+      dockerfile: openwisp_freeradius/Dockerfile
+    deploy:
+      replicas: 1
+    env_file:
+      - .env
+    depends_on:
+      - postgres
+      - dashboard
```
2stacks: I think this should be radius container.

kubernetes/Service.yml
```diff
@@ -91,6 +91,24 @@ spec:
 ---
 apiVersion: v1
 kind: Service
+metadata:
+  name: freeradius
+  labels:
+    app: freeradius
+spec:
+  type: ClusterIP
```
2stacks: Need to think about how the Freeradius service will be reachable outside the cluster.  First thought is make the service Type = 'NodePort' but using 'externalIPs' may be enough.  Also, any multi-node cluster will need some external load balancing and most of the public clouds do not support UDP in their elastic load balancers.

build/openwisp_freeradius/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    DB_HOST=postgres \
+    DB_PORT=5432 \
+    DB_USER=admin \
+    DB_PASSWORD=admin \
+    DB_DATABASE=openwisp_db \
+    DB_SSLMODE=disable \
```
2stacks: Did we decide to drop database SSL as an option?  I know we talked about it.  If so then `DB_SSLMODE` option can be removed.

kubernetes/Service.yml
atb00ker: Thanks for informing, I did not know about Loadbalancers not supporting [UDP for most providers](https://kubernetes.io/docs/concepts/services-networking/service/#udp).

docker-compose.yml
atb00ker: As per my understanding,since both the radius and dashboard container have the radius API, so I think freeradius container should work with both dashboard and radius containers. The default one that I am testing on is using dashboard, hence I have put dashboard here.

If a user is creating a new instance of openwisp-radius container, they'll have to create a freeradius container working with it. (I think I should add this in the documentation.)
However, since the openwisp-radius container is optional while openwisp-dashboard is always going to be there, I thought that I should keep dashboard as default here. 

Please let me know your thoughts on this. :smile:

build/openwisp_freeradius/Dockerfile
atb00ker: Oh yes, the name here makes it all confusing, I am changing the variable name to `RADIUS_API_DOMAIN`.
Since, Freeradius can work with both dashboard and radius and I am using dashboard for testing, that's why it was dashboard here.
Yes, the traffic is going through nginx because the containers are running `uwsgi` which is unsupported protocol for freeradius.

build/openwisp_freeradius/Dockerfile
atb00ker:
```
    export DB_SSLMODE=`echo "$DB_OPTIONS" | jq -r '.sslmode'`
    if [ "$DB_SSLMODE" == "null" ]; then
        export DB_SSLMODE=disable
    fi
```

The `DB_OPTIONS` is a json variable taking care of additional database variables, I used that to take care about the SSL settings with database.
(Are you reffering to the discussion here: https://groups.google.com/d/msg/openwisp/Mh0tMBrB2Js/0cXVIr62AQAJ or some other discussion?)

docker-compose.yml
2stacks: I thought the purpose of the openwisp-radius container was for serving the Radius API.  I didn't think we were going to leave the Radius API urls as part of dashboard.  I may have misunderstood so we should at least have documentation that makes it less confusing.

build/openwisp_freeradius/Dockerfile
2stacks: Thanks, I missed that the variables had moved to DB_OPTIONS.

.env
```diff
@@ -39,6 +40,11 @@ POSTFIX_RELAYHOST_TLS_LEVEL=may
 POSTFIX_RELAYHOST=False
 POSTFIX_RELAYHOST_USERNAME=False
 POSTFIX_RELAYHOST_PASSWORD=False
+# Radius
+FREERADIUS_ORGANIZATION=default
+FREERADIUS_TOKEN=djangofreeradiusapitoken
```
nemesisdesign: why are the previous two variables needed?
In openwisp-radius the token is managed at organization level in the DB (from the admin interface).

build/openwisp_dashboard/load_init_data.py
```diff
@@ -13,5 +15,44 @@
...
+def set_default_radius_token(radiusOrg):
+    '''
+    Set FREERADIUS_TOKEN to the given organization.
+    '''
+    radiusConf = OrganizationRadiusSettings.objects \
+                                           .filter(organization=radiusOrg)
+    if not radiusConf.exists():
+        radiusConf = OrganizationRadiusSettings()
+        radiusConf.organization = radiusOrg
+        radiusConf.token = os.environ['FREERADIUS_TOKEN']
```
nemesisdesign: why don't we just let the model generate the token?
It feels weird to have an env var for something that is then managed in the DB

build/openwisp_freeradius/raddb/mods-available/sqlcounter
```diff
@@ -0,0 +1,41 @@
...
+   counter_name = Max-Daily-Session-Traffic
+   check_name = Max-Daily-Session-Traffic
+   sql_module_instance = sql
+   key = 'User-Name'
+   reset = daily
+   query = "SELECT SUM(acctinputoctets + acctoutputoctets) \
+            FROM radacct \
+            WHERE UserName='%{${key}}' \
+            AND UNIX_TIMESTAMP(acctstarttime) + acctsessiontime > '%%b'"
+}
```
nemesisdesign: what happens if we want to modify any of these freeradius configuration files? Can we do it without altering the source code of the image?

build/openwisp_freeradius/raddb/sites-enabled/default
```diff
@@ -0,0 +1,59 @@
...
+
+        Post-Auth-Type REJECT {
+            update control { &REST-HTTP-Header += "${....api_token_header}" }
+            rest
+        }
+    }
+
+    pre-proxy {}
+    post-proxy {}
+}
```
nemesisdesign: this one for example, not only may vary from user to user, but each organization may need more than one site.
the thing is that this is more configuration than code, I'm not sure if it's right to include this in the image, from a certain point of view we do need to include something in the image if we want it to work out of the box, but I worry that if we don't provide a way to override these freeradius configurations, the project will not be as useful as we want it to be

docker-compose.yml
nemesisdesign: APIs in general should not be part of the dashboard

build/openwisp_freeradius/raddb/mods-available/sqlcounter
2stacks: I think you would have to mount over the raddb directory or mount individual files as needed.

.env
2stacks: The freeradius rest module expects the token to be hardcoded or available via the env.  The service wont start if its missing.

docker-compose.yml
atb00ker: Got it, thanks, opening a seperate issue to remove API from dashboard container. #52

kubernetes/Service.yml
atb00ker: Closing this as per the conversation on Gitter regarding all major providers supporting UDP now! :smile:

.env
atb00ker: Changed this part. Now, organization name needs to be passed to container which checks if the given org has radius_token set from database, if yes, it fetches it otherwise, container exits with code 1.

build/openwisp_dashboard/load_init_data.py
atb00ker: Removed this, It was used previously, not in use anymore! :smile:

build/openwisp_freeradius/raddb/mods-available/sqlcounter
atb00ker: I agree, the default rabbd/ files are just sample of basic configuration for testing (and can be used as an example for user), user can mount their own raddb directory for customization. :smile:

build/openwisp_freeradius/raddb/sites-enabled/default
atb00ker: The default `rabbd/` files are just sample of basic configuration for testing (and can be used as an example for user), user can mount their own raddb directory for customization. :smile:

2stacks:
We should add to documentation how a user can mount config directory/files
in both docker and Kubernetes.

build/openwisp_freeradius/raddb/mods-available/sqlcounter
atb00ker: Added documentation for the same. :smile:

atb00ker:
> We should add to documentation how a user can mount config directory/files
in both docker and Kubernetes.

Done! :+1:

build/openwisp_freeradius/raddb/mods-available/sqlcounter
2stacks: Check the docs for Freeradius/Kubernetes, it looks like it has some copy paste from the Nginx section.

atb00ker: Oh, thanks! Fixed it! :smile:

build/openwisp_nginx/nginx.template.conf
```diff
@@ -31,4 +30,20 @@ http {
     server_names_hash_bucket_size $NGINX_SERVER_NAME_HASH_BUCKET;
 
     include /etc/nginx/conf.d/*.conf;
+
+        server {
```
nemesisdesign: fix the indentation of this block of configuration please

docs/FILES.md
```diff
@@ -0,0 +1,104 @@
+# Configuration Files
```
nemesisdesign: :+1:

PR #46, opened 2019-07-15
[enchancement] Add OpenVPN container

### LINKED ISSUE #18: [tasklist] OpenVPN image
> Provide a default management VPN based on OpenVPN in a dedicated container. The VPN shall be also pre-configured in OpenWISP, both as a VPN Server and a related VPN client template (ask questions to your mentors to know more about this), the OpenVPN configuration used in the container shall be kept in sync with the definition available in the VPN Server of OpenWISP. The Certificate Revocation List of the VPN shall be downloaded periodically from openwisp on the filesystem (a script run in a crontab) of the VPN server and the configuration of the VPN server

atb00ker:
Closes #18


build/openwisp_dashboard/load_init_data.py
```diff
@@ -2,16 +2,113 @@
...
+    needs to be made
+    '''
+    return Organization.objects.get(slug=vpn_org)
+
+
+def create_default_CA(vpnOrg):
+    '''
+    Create default certificate authority
+    '''
+    if not Ca.objects.filter(name='default').exists():
```
nemesisdesign: what happens if a user renames the `default` VPN to something else?
will a new VPN be created?

build/openwisp_openvpn/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM kylemanna/openvpn:latest
+
+RUN apk add --no-cache postgresql-client supervisor && \
```
nemesisdesign: why do we need `posgresql-client`?

build/openwisp_openvpn/openvpn.sh
```diff
@@ -0,0 +1,14 @@
...
+# update OpenVPN configurations periodically.
+source /utils.sh
+
+init_conf
+openvpn_config
+openvpn_config_checksum
+
+if [ "${OFILE}" != "${NFILE}" ]; then
+    openvpn_config_download
+    supervisorctl restart openvpn
```
nemesisdesign: why was supervisor chosen to run OpenVPN?

atb00ker: Supervisor is used because I need to restart OpenVPN after crl list is updated or configurations are changed.
If I use OpenVPN as the service keeping the docker container running, restarting would mean killing the container while supervisor helps me prevent that!

build/openwisp_openvpn/Dockerfile
atb00ker: I am using `psql` command in this container and that is asking for postgresql-client.
The command I am using is:
```
psql -qAtc "SELECT id,key FROM config_vpn where name='${VPN_NAME}';"
```
This gets me the `UUID` and `KEY` of the vpn which I use for fetching files from API as:
```
wget -qO vpn.tar.gz --no-check-certificate \
    ${DOMAIN}/controller/vpn/download-config/$UUID/?key=$KEY
```

build/openwisp_dashboard/load_init_data.py
atb00ker: Renaming VPN Server name will not cause any problem, just a new instance of OpenVPN server with the new VPN name will need to be created. ( i.e the environment variable `VPN_NAME` needs to be manually updated in the VPN server instance.)
If you rename the certificate or CA associated (that are name `default`), then everything will keep working except when dashboard instance is restarted for any reason, a CA named default and a Cert named default will be created, both useless but other than that,  everything will keep working.

build/openwisp_openvpn/openvpn.sh
nemesisdesign: great, can you add this explanation somewhere as a comment please?

build/openwisp_openvpn/Dockerfile
nemesisdesign: I see. This is related to the comment above. We may need to think about a better solution for the default VPN in the future. For the moment let's keep it as is and let's go ahead.

atb00ker:
> what's the status of this one?
Waiting for #44 to be merged! :smile:

build/openwisp_openvpn/openvpn.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# This script will be called by cronjob to
+# update OpenVPN configurations periodically.
+source /utils.sh
+
+init_conf
```
2stacks: Does this need/is it safe to run init_conf periodically?  I assume this is to capture a TZ or Database password change?

build/openwisp_openvpn/revokelist.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
2stacks: Since you are using checksums to validate if configs have changed would it be more maintainable to combine revokelist.sh and openvpn.sh?  Just have a single script that runs via cron to perform routine maintenance and configuration update?

atb00ker: I kept them in different scripts because I assume that the frequency of updating revoke list would much less, because after running the revoke list script, openvpn is restarted, so revoke list is updated only at 3AM on Sundays, while openvpn config is checked every minute because I assume the user will want to see the updated settings as soon as possible! 
Is my understanding about the frequiency of scripts correct? :smile:

build/openwisp_openvpn/openvpn.sh
atb00ker: It's safe to run, It's only settings the PSQL default variables, but I can see this being a bit confusing in future, in case someone adds something to `init_conf` which isn't safe to run again.

So, I'll make psql a seperate function and make it more clear by adding a comment! :smile:

atb00ker: Seperated the default psql variables in a new function now! :smile:

build/openwisp_openvpn/revokelist.sh
2stacks: I assumed you had a good answer :smile:  Thanks!

2stacks: Is openvpn required to restart just from checking for revocations or only when it needs to update it's list of revoked certificates?  It seems like it should check daily and only restart as necessary.

build/openwisp_openvpn/openvpn.sh
atb00ker: Done! :+1:

build/openwisp_openvpn/revokelist.sh
atb00ker: Okay, I am not very sure about how CRLs work, but everytime I download the CRL, I get a new file with different SHA1, so I assumed that that's expected nature of CRL file. Hence I can't know for sure if there is a change in revokation list and therefore I need to restart OpenVPN even after just downloading the list (with optentially no change.)

i.e. If I download `http://dashboard.openwisp.org/x509/ca/1.crl` and then download the same again, I get different values of the crl file.

Issue #8, opened 2019-05-20
[tasklist] celery image


atb00ker:
> Provide a celery worker, initially used only by OpenWISP Controller (connections branch), but which later will executes background tasks of other modules as well, in a dedicated container.

source: http://openwisp.org/gsoc/ideas-2019.html

**Research work:**
- https://nick.sarbicki.com/blog/deploy-django-celery-on-k8s/ 

~Blocked by: #40~

PR #48, opened 2019-07-23
[enhancement] Add celery image

atb00ker:
Closes: #8

### DIFF
--- a/build/common/init_command.sh
+++ b/build/common/init_command.sh
@@ -45,6 +45,9 @@ elif [ "$MODULE_NAME" = 'nginx' ]; then
         envsubst_create_config /etc/nginx/openwisp.template.conf http
     fi
     nginx -g 'daemon off;'
+elif [ "$MODULE_NAME" = 'celery' ]; then
+    python services.py database redis dashboard
+    celery -A openwisp worker -l info
 else
     python services.py database redis dashboard
     start_uwsgi
--- a/tests/tests.sh
+++ b/tests/tests.sh
@@ -6,6 +6,7 @@ This script will ensure that the following services are running:
 - Postfix
 - Websocket
 - Freeradius
+- Celery
 """
 
 function print_services_logs {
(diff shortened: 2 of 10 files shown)


tests/tests.sh
```diff
@@ -19,73 +20,84 @@ function print_services_logs {
...
 function test_websocket {
     # This test ensures that wesocket service is running
     # correctly by trying to reach \ws location in nginx.
-    timeout 2 $CURL_BIN --include \
+    $CURL_BIN -I -X GET --include \
                         --no-buffer \
                         --header "Connection: Upgrade" \
                         --header "Upgrade: websocket" \
                         --header "Host: dashboard.openwisp.org" \
                         --header "Origin: https://dashboard.openwisp.org" \
```
2stacks: Looks like you removed most of the hard coding of 'dashboard.openwisp.org' except for here.  Shouldn't matter for controlled tests but it seems inconsistent.

nemesisdesign: :+1:

build/common/init_command.sh
```diff
@@ -37,6 +37,9 @@ elif [ "$MODULE_NAME" = 'nginx' ]; then
         envsubst_create_config /etc/nginx/openwisp.template.conf http
         nginx -g 'daemon off;'
     fi
+elif [ "$MODULE_NAME" = 'celery' ]; then
+    python services.py database redis dashboard
+    celery -A openwisp worker -l info
```
atb00ker: > are we sure that this service does not expose URLs right?

This container is not exposed outside the cluster! :smile: 

> It should only run the celery worker, but I don't see where the celery worker is being launched.
Can you point it out please?

When  `"$MODULE_NAME" = 'celery'` -> `celery -A openwisp worker -l info`

tests/tests.sh
atb00ker: I am not too worried about this file because the shell script tests have failed me once or twice now, so I am moving tests to selenium. 
However, I have fixed this for now! :smile:

PR #66, opened 2019-10-12
[bug] Fix postfix relay access error

atb00ker:
Changed `smtpd_recipient_restrictions` to fix relay access denied issue.


docs/ENV.md
```diff
@@ -269,19 +269,19 @@ Additionally, you can search for the following:
 ### `POSTFIX_RELAYHOST`
 - **Explaination:** Host that relays your mails.
 - **Valid Values:** IP address | Domain
-- **Default:** --BLANK--
-- **Example:** smtp.gmail.com:587
+- **Default:** False
```
2stacks: As disused, using False for an ENV that doesn't require only True|False seems confusing.  Recommend using null or None since defaults values are required by Docker and can not be left blank.

atb00ker: Done! :smile:

2stacks:
I think there's still an issue with GDAL that prevents the dashboard
container from starting.

atb00ker:
Merging this, it's approved and the tests fail because of reasons not related to this pull request! :smile:

PR #67, opened 2019-10-12
[deployment] Improved kubernetes deployment

atb00ker:
Changes inspired by deploying on a bare-metal cluster and GKE cluster:
- Add GKE deployment compatibility
- Improved kubernetes documentation
- Simplified documentation
- Added openwisp-nfs server for kubernetes deployments
- Simplified & fixed makefile circular calls
- Other small fixes/improvements


docs/GCP.md
```diff
@@ -0,0 +1,38 @@
...
+1. External IP: Create a global static IP and 2 IP address in the same region as you want to create your cluster in.
+2. Configure your domain with following A records:
+
+    ```
+    dashboard.openwisp.org      GLOBAL-IP
+    controller.openwisp.org     GLOBAL-IP
+    radius.openwisp.org         GLOBAL-IP
+    topology.openwisp.org       GLOBAL-IP
+    freeradius.openwisp.org     REGIONAL-IP-1
+    openvpn.openwisp.org        REGIONAL-IP-2
```
2stacks: Maybe we should use a place holder or variable like `dashboard.<your.domain>`.

kubernetes/ExtServices.yml
```diff
@@ -0,0 +1,61 @@
...
+      - backend:
+          serviceName: openwisp-nginx
+          servicePort: 80
+        path: /*
+  - host: controller.openwisp.org
+    http: *https_rules
+  - host: radius.openwisp.org
+    http: *https_rules
+  - host: topology.openwisp.org
+    http: *https_rules
```
2stacks: Looks like this is in both ExtServices and Ingress.

README.md
```diff
@@ -41,17 +41,21 @@ The configuration for openwisp images using environment variables is available [
...
-2. Changes for your cluster:
+1. Setup a Kubernetes Cluster
+
+   1.1. A guide for setting up the cluster on bare-metal machines is available [here](https://blog.alexellis.io/kubernetes-in-10-minutes/) and the guide to get started with kubernetes-dashboard (Web UI) is available [here](https://github.com/kubernetes/dashboard).
 
-   2.1. `externalIPs` in `Service.yml` should to be your cluster's `externalIPs`
+   1.2. On providers you may initiate a cluster as per your desired configurations. (Read the [GCP cluster requirement](docs/GCP.md) to match those requirements in your cluster)
 
-   2.2 `ingress` in `Ingress.yml` should to be your cluster's loadbalancer IPs.
+2. Changes for your cluster:
```
2stacks: Need to add info regarding updating all domain names.  Domain names currently need to be configured in Ingress.yml, ExtServices.yml, Configmap.yml, Certificates.yml.  Perhaps we should make all other files pull from the Configmap.yml so that domain names only need to be set in one place.

docs/GCP.md
atb00ker: Yes, that sounds much better to avoid any confusion. :smile: 
On it. :+1:

kubernetes/ExtServices.yml
atb00ker: Yes, when I apply the "Ingress.yml" file on Google Cloud, it doesn't work and keeps complaining.
So, what I had to do is, create "ExtServices.yml" which doesn't work, it only creates an ingress and defines the path, then, I apply the "Certificates.yml" which gets me the letsencrypt certificates for me,
Finally, I apply the "Ingress.yml" which has the same block with
```
  tls:
  - hosts:
        ...
       (hosts)
        ...
    secretName: openwisp-2stacks-crt
``` 
So that I can finally use the site with https! :smile:

2stacks:
Sounds like an order of operations problem.  Feels like we should log this
as an issue.

atb00ker:
> Sounds like an order of operations problem.  Feels like we should log this
as an issue.

It kinda makes sense that it wouldn't work.
The `secret: openwisp-crt` is created by the "Certificate.yml" upon successful obtaining of certificate and ingress needs that secret file. So, if the secret doesn't exist, I suppose we should expect it to throw errors.

kubernetes/ExtServices.yml
2stacks: I see what you mean with the circular references.  It seems like the step should be to;

1. Create the ingress without tls specification.
2. Create the certificate
3. Update the ingress with the tls specification.

That's pretty much what you're doing I guess but wonder if there's a way without duplicating so many lines of yaml.

atb00ker: Indeed, that'll be ideal. Looking into the cert-manager docs for the same. :smile:

atb00ker:
Update: Still working on it, fixed the issue, need to improve documentation last commit: https://github.com/atb00ker/docker-openwisp/commit/1084ff753a3c7cc3ec51dda0e2dac80f3cd8c32b

kubernetes/ExtServices.yml
atb00ker: Solved! :smile:

atb00ker:
Changes inspired by deploying on a bare-metal cluster and GKE cluster:
- Add GKE deployment compatibility
- Improved kubernetes documentation
- Simplified documentation
- Added openwisp-nfs server for kubernetes deployments
- Simplified & fixed makefile circular calls
- Other small fixes/improvements

atb00ker:
> Sounds like an order of operations problem.  Feels like we should log this
as an issue.

@2stacks I tired a lot of things, but that there is no fix for that atleast right now.

However, I have managed to reduced the repeated code:
Before cert-manager:
```
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: openwisp-http-ingress
  annotations:
    kubernetes.io/ingress.global-static-ip-name: openwisp-http-loadbalancer-ip
  labels:
    app: openwisp-nginx
spec:
  backend:
    serviceName: openwisp-nginx
    servicePort: 80
```

After cert-manager gets certificates from this ingress:

```
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: openwisp-http-ingress
  annotations:
    kubernetes.io/ingress.global-static-ip-name: openwisp-http-loadbalancer-ip
  labels:
    app: openwisp-nginx
spec:
  backend:
    serviceName: openwisp-nginx
    servicePort: 80
  tls:
  - hosts:
    - dashboard.openwisp.org
    - controller.openwisp.org
    - radius.openwisp.org
    - topology.openwisp.org
    secretName: openwisp-tls-secret
```

atb00ker:
Ah, I'll have to change ingress again and add :
```
  rules:
  - http:
      paths:
      - backend:
          serviceName: openwisp-nginx
          servicePort: 80
```

kubernetes/ExtServices.yml
atb00ker: Un-solve, the problem is that ingress-gce doesn't let ingress work if it has `tls` and no certificates.
So I had to a patch:
```
kubectl patch ingress/openwisp-http-ingress \
    --patch '{
        "spec": {
            "tls": [
                {
                "hosts": [
                    "dasboard.openwisp.org",
                    "controller.openwisp.org",
                    "radius.openwisp.org",
                    "topology.openwisp.org"
                ],
                "secretName": "openwisp-tls-secret"
                }
            ]
        }
    }'
```

There may be fix for this in future, so creating an issue

PR #75, opened 2019-12-21
[chores] Minor fixes & openwisp-utils-qa-checks

atb00ker:
- Added openwisp-utils check
- Removed port 25 from exposed ports of dashboard


atb00ker:
Will continue to work on this after the GCI PRs are merged because I don't want GCI students to have troubles with flak8 changes.

build/common/openwisp/tasks.py
```diff
@@ -17,10 +17,12 @@ def radius_tasks():
     management.call_command("delete_old_users",
                             older_than_months=int(os.environ['CRON_DELETE_OLD_USERS']))
 
+
 @shared_task
 def save_snapshot():
     management.call_command("save_snapshot")
 
+
 @shared_task
```
nemesisdesign: I thought that this code could now be added to openwisp-network-topology

atb00ker: I'll investigate this ! :smile:

atb00ker: If I move celery commands from their own container to the container of the application(example: network-topology container), then the network-topology container would have to run both the celery task and the uwsgi server, which is not inline with the docker philosophy of running one thing in one container, what do you suggest?

PR #79, opened 2019-12-30
[bug] Fix installation & debian changes

atb00ker:
- Updated according to new version of django-freeradius
- Version lock for redis to avoid websocket bug
- Removed pip installation from git (It's not required!)

### DIFF
--- a/build/openwisp_base/Dockerfile
+++ b/build/openwisp_base/Dockerfile
@@ -3,8 +3,10 @@ FROM python:3.7.5-slim-buster AS SYSTEM
 # System requirements:
 # 1. Pillow: zlib1g-dev libjpeg-dev
 # 2. gettext: Required by envsubst used in scripts
+# 3. django-freeradius/weasyprint: libcairo2 libpangocairo-1.0
 RUN apt update && \
-    apt install --yes --no-install-recommends zlib1g-dev libjpeg-dev gdal-bin gettext && \
+    apt install --yes --no-install-recommends zlib1g-dev libjpeg-dev \
+                libcairo2 libpangocairo-1.0 gdal-bin gettext && \
     rm -rf /var/lib/apt/lists/* /root/.cache/pip/* /tmp/*
 
 FROM SYSTEM AS PYTHON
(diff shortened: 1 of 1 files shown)


build/openwisp_base/Dockerfile
```diff
@@ -13,33 +15,32 @@ FROM SYSTEM AS PYTHON
...
 RUN pip install --upgrade https://github.com/openwisp/django-netjsonconfig/tarball/master \
                 --prefix='/install'
 RUN pip install --upgrade https://github.com/openwisp/django-x509/tarball/master \
                 --prefix='/install'
-RUN pip install --editable git+git://github.com/openwisp/openwisp-utils#egg=openwisp_utils \
+RUN pip install --upgrade https://github.com/openwisp/openwisp-utils/tarball/master \
                 --prefix='/install'
 RUN pip install --prefix='/install' asgi_redis service_identity django-redis psycopg2 uwsgi \
-                                    sentry-sdk supervisor celery django-celery-beat redis
+                                    sentry-sdk supervisor celery django-celery-beat redis==2.10.6
```
atb00ker: For context: redis is set to 2.10.6 because v3 is not compatible with django-channels.

build/openwisp_base/Dockerfile
```diff
@@ -13,33 +15,32 @@ FROM SYSTEM AS PYTHON
 # 1. Twisted: gcc
 # 2. Postgresql: libpq-dev
 RUN apt update && \
-    apt install --yes --no-install-recommends git gcc libpq-dev && \
+    apt install --yes --no-install-recommends gcc libpq-dev && \
```
atb00ker: Removed git because it's not required and `git+` installation seem to take more time.

build/openwisp_base/Dockerfile
```diff
@@ -13,33 +15,32 @@ FROM SYSTEM AS PYTHON
...
+RUN pip install --upgrade https://github.com/openwisp/openwisp-controller/tarball/master \
                 --prefix='/install'
-RUN pip install --editable git+git://github.com/openwisp/openwisp-radius#egg=openwisp_radius \
-                --prefix='/install'
-RUN pip install --editable git+git://github.com/openwisp/openwisp-network-topology#egg=openwisp_network_topology \
+RUN pip install --upgrade https://github.com/openwisp/openwisp-network-topology/tarball/master \
                 --prefix='/install'
 RUN pip install --upgrade https://github.com/openwisp/django-freeradius/tarball/master \
                 --prefix='/install'
+RUN pip install --upgrade https://github.com/openwisp/openwisp-radius/tarball/master \
```
atb00ker: `openwisp-radius` is moved below because of an on going issue in `openwisp-radius` that needs to be fixed seperately: https://github.com/openwisp/openwisp-radius/issues/73

This is only a quick fix.

PR #81, opened 2020-01-04
[tests] Tested websocket using selenium

oltarzewskik:


oltarzewskik:
Build is failing even if I will send PR with zero changes (https://travis-ci.org/openwisp/docker-openwisp/builds/632750224?utm_source=github_status&utm_medium=notification) 

however please review my code and I have few questions:
- Should I delete location after tests?
- Should I check if there is no other location with name `test-location`, in `travis` this situation won't happen however while testing on local machine it potentially can happen

And I decided to not use `unittest` python module because I think we can go without it

tests/selenium_test_websocket.py
```diff
@@ -0,0 +1,66 @@
+from selenium import webdriver
+
+def log_in(driver):
```
atb00ker: You know, these functions are really common and can be used for testing other features in future? 
Can you please make class and put all the common functions and use from it? :smile:

tests/selenium_test_websocket.py
```diff
@@ -0,0 +1,66 @@
...
+    if not 'admin/login' in driver.current_url:
+        return
+    username_input = driver.find_element_by_name('username')
+    password_input = driver.find_element_by_name('password')
+    username_input.send_keys('admin')
+    password_input.send_keys('admin')
+    log_in_button = driver.find_element_by_class_name('submit-row').find_element_by_xpath('//input[@type="submit"]')
+    log_in_button.click()
+
+def create_location(driver):
```
atb00ker: Can you please make a another class to put all the websocket related tests together, just for the sake of keeping stuff together! :smile:

tests/tests.sh
```diff
@@ -99,6 +99,11 @@ function test_websocket {
                "(You may want to increase test timeout)"; FAILURE=1; }
 }
 
+function selenium_test_websocket {
+    pip install selenium
```
atb00ker: `pip install selenium` is a one-off job / management job.
Please move it `.travis.yml` it shouldn't run everytime this function is called! :smile:

tests/tests.sh
```diff
@@ -99,6 +99,11 @@ function test_websocket {
...
+    python selenium_test_websocket.py
```
atb00ker: If this fails, I hope travis jobs also fails, I'll have to test it! :smile:

atb00ker:
P.S: This repository doesn't have flake8 yet, can you please manually make sure your code is pep8! :smile:

Flake8 would be added soon! :smile:

oltarzewskik:
@atb00ker I made these code as reusable as possible and checks as pragamtic as possible
if you have any questions to code or you don't understand something (for example `delete_location` function) please tell me :smile: 

Code is pep8 consistent :smile:

tests/selenium_tests.py
```diff
@@ -0,0 +1,135 @@
...
+            delete_button.click()
+            confirm_delete_button = (
+                driver.
+                find_element_by_xpath('//div/div[@id="content"]/form/div/input[@type="submit"]')
+            )
+            confirm_delete_button.click()
+
+
+class TestWebSocket(TestUtilities, unittest.TestCase):
+    base_driver = webdriver.Chrome()
```
atb00ker: Can we please have an option to have Chrome headless?
Like if an arguement `--headless` is passed to python scripts, it starts all the chrome instances as  headless.
Also, please enable headless on `.travis.yml`

tests/selenium_tests.py
```diff
@@ -0,0 +1,135 @@
...
+    remove_locations = True
+    js_errors = []
+
+    def tearDown(self):
+        if self.remove_locations:
+            self.delete_location(self.base_driver, self.location_name)
+        self.base_driver.close()
+        js_websocket_errors = []
+        for error in self.js_errors:
+            if 'websocket' or 'ws' in error.lower():
```
atb00ker: Why is this in tearDown() ? This is a part of the test itself and should be inside the `test_websocket()`, remember teadDown is a special function to remove tests data. Google about it! :smile:

tests/selenium_tests.py
```diff
@@ -0,0 +1,135 @@
...
+            confirm_delete_button = (
+                driver.
+                find_element_by_xpath('//div/div[@id="content"]/form/div/input[@type="submit"]')
+            )
+            confirm_delete_button.click()
+
+
+class TestWebSocket(TestUtilities, unittest.TestCase):
+    base_driver = webdriver.Chrome()
+    location_name = 'test-location'
```
atb00ker: This can have a name that users usually don't use, like `automated-websocket-selenium-loc01`

tests/selenium_tests.py
```diff
@@ -0,0 +1,135 @@
...
+    remove_locations = True
```
atb00ker: I am sure thought something here, but I am not sure if I get it, can you please tell when do you not what to delete the instance that we created during the test!? :smile:

tests/selenium_tests.py
oltarzewskik: I wanted to return js errors even if some checks fail, it is reason why I put it here.

tests/selenium_tests.py
oltarzewskik: `delete_location` deletes all locations with given name therefore if location(s) with given name already exists we don't want to delete it instead return only fail with message. It is only case when don't want to delete the location :smile:

tests/selenium_tests.py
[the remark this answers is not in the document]
oltarzewskik: I will try :smile:

tests/selenium_tests.py
atb00ker: Why wouldn't it run in the test_websocket function? :smile:
If `try`, `except` is used properly at places where failure can occure, it will check for JavaScript errors even on failure.
Infact this check can be before the location is saved via the second chrome instance.

oltarzewskik:
@atb00ker I made changes it looks like `--headless` is enabled on `travis` by default, however I stuck on adding `--headless` attribute while running python script.

This code will work:
```
import sys
import argparse
import unittest

if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('--headless', action='store_true')
    args = parser.parse_args()
    print(args)
```
this won't:
```
import sys
import argparse
import unittest

if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('--headless', action='store_true')
    args = parser.parse_args()
    print(args)
    unittest.main()
```

atb00ker:
@oltarzewskik can you please try something like this: https://stackoverflow.com/a/20702984/6410464 ?

oltarzewskik:
@atb00ker I was trying this I was trying "[update] " section, because I think we don't want use first
`TestWebSocket.headless = os.environ.get(args.headless, TestWebSocket.headless)`
will give error: `TypeError: str expected, not bool`, because `--headless` is `bool` and work like switch.

I was trying doing something like that:
`TestWebSocket.headless = os.environ.get(str(args.headless), TestWebSocket.headless)` and later on validate is it `bool` `True` or `False` however when I pass `python test.py --headless` I get error as described in comment on top: `error: unrecognized arguments: --headless`

atb00ker:
I see `python: can't open file 'selenium_tests.py': [Errno 2] No such file or directory` and the travis passes, please make sure travis fails in such cases! :smile:

.travis.yml
```diff
@@ -1,14 +1,26 @@
...
+
 branches:
   only:
   - master
   - travis
 
+before_script:
+  - sudo apt-get update
+  - sudo apt-get install google-chrome-stable=73.0.3683.86-1 -y
+  - wget https://chromedriver.storage.googleapis.com/73.0.3683.68/chromedriver_linux64.zip
```
atb00ker: Any specific reason you needed to move to chrome 73 ? :smile:

oltarzewskik:
> Regarding your issue with passing headless variable, can you share why do we not want to use the mentioned comment before "Updated" ?

It is not that much scalable, if we in future will add other argument and run this like that: `python selenium_tests.py headless --something` it won't work. However now I think that we can simply assert that is `--headless` in `sys.argv` which will be more scalable :smile:

.travis.yml
oltarzewskik: I am actually not moving backward to chrome 73, because it is newest available version of `google-chrome-stable` on this travis. I only added check that that version is installed, because it will work with `chromedriver` :smile:

tests/tests.sh
```diff
@@ -82,21 +82,8 @@ function test_freeradius {
...
-                        --header "Upgrade: websocket" \
-                        --header "Origin: ${APP_URL}" \
-                        --header "Sec-WebSocket-Key: x3JJHMbDL1EzLkh9GBhXDw==" \
-                        --header "Sec-WebSocket-Version: 13" \
-               ${APP_URL}/ws/ | \
-        grep -q "101 Switching Protocols" && \
-        { echo "SUCCESS: Websocket service is working!"; } || \
-        { echo "ERROR: Websocket service did not respond!" \
-               "(You may want to increase test timeout)"; FAILURE=1; }
+function selenium_tests {
```
atb00ker: The old name `test_websocket` was fine, please revert the name! :smile:

tests/selenium_tests.py
```diff
@@ -0,0 +1,122 @@
...
+        self.set_location_marker(change_location_driver, self.location_name)
+        change_location_driver.close()
+        # assert that marker popped up on map
+        self.assertEqual(len(self.base_driver.find_elements_by_class_name('leaflet-marker-icon')), 1)
+
+
+if __name__ == '__main__':
+    # To run tests in chrome headless mode pass --headless command line argument
+    if '--headless' in sys.argv:
+        sys.argv.remove('--headless')
```
atb00ker: Don't remove it. 
Might come in handly somewhere else in future! :smile:

oltarzewskik: If I won't delete it I get `error: unrecognized arguments: --headless` error, popping it also removes it

atb00ker: Okay. I'll look into it later sometime! :smile:

atb00ker:
Not sure if that's going to work either:
```
python selenium_tests.py --headless --something
```

I think both `headless` and `--headless`work the same way in this context! :smile:

atb00ker:
Okay, I investigated.
The root cause of the issue is : we need to remove the arguements because unitest.TestProgram has it's own instance of argparser which is not meant to be inherited and is conflicting with the arguements you pass from command line! :smile:

tests/tests.sh
```diff
@@ -158,4 +144,5 @@ function init_tests {
     if [[ $FAILURE = 1 ]]; then
         exit $FAILURE
     fi
+    test_websocket
```
atb00ker: You moved it after script exits? It'll never run. Well that's certainly not what I meant when I said make it the last test! :joy:

tests/selenium_tests.py
atb00ker: You are using `sys.argv` both `headless` and `--headless` do the same job, I think you wanted to use `argparser` in which `--headless` is a optional arguement with a possible value like `--headless True`.
In anycase, I tired argparser today, I can't work without a ugly hack here, so I dropped the idea and moved to a json file. :smile:

tests/selenium_tests.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def setUp(self):
+        self.chrome_options = Options()
+        if self.is_headless:
+            self.chrome_options.add_argument("--headless")
+        self.base_driver = webdriver.Chrome(options=self.chrome_options)
+        self.location_name = 'automated-websocket-selenium-loc01'
+
+    def tearDown(self):
+        self.delete_location(self.base_driver, self.location_name)
```
atb00ker: Fun fact: I learned today that if tests fail, `tearDown` is not called in unittest, so, I removed it! :smile:

PR #84, opened 2020-01-07
[chores] Temporary version locking

atb00ker:
Did temporary version locking to avoid
the containers failing every once in a while
because of update in some other repository

### DIFF
--- a/build/openwisp_base/Dockerfile
+++ b/build/openwisp_base/Dockerfile
@@ -20,24 +20,23 @@ RUN apt update && \
 
 ENV PYTHONPATH=/install/lib/python3.7/site-packages
 # Python Packages
-RUN pip install --upgrade https://github.com/openwisp/openwisp-controller/tarball/master \
-                --prefix='/install'
-RUN pip install --upgrade https://github.com/openwisp/openwisp-network-topology/tarball/master \
-                --prefix='/install'
-RUN pip install --upgrade https://github.com/openwisp/django-freeradius/tarball/master \
-                --prefix='/install'
-RUN pip install --upgrade https://github.com/openwisp/openwisp-radius/tarball/master \
-                --prefix='/install'
-RUN pip install --upgrade https://github.com/openwisp/openwisp-users/tarball/master \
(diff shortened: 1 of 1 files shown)


atb00ker:
> What about using git tags for those modules which have been released recently?

I thought, I'd do them all together when the other modules are released but sure, I'll check which ones are released and use git tags.

> Also, it could be interesting to make this configurable, because a developer may want to install a customized version (or a development version, or a fork) of one or more modules.

Can be done using build args, although I think the number of steps to make the changes would remain the same, so it wouldn't make it any easier for someone building their own images!

> I have on my todo list to release also the remaining ones (django-netjsonconfig, openwisp-controller, django-netjsongraph, openwisp-network-topology, django-freeradius, openwisp-radius).

That's great, waiting for it! :smile:

atb00ker:
>> Also, it could be interesting to make this configurable, because a developer may want to install a customized version (or a development version, or a fork) of one or more modules.

> Can be done using build args, although I think the number of steps to make the changes would remain the same, so it wouldn't make it any easier for someone building their own images!

I found a problem with this. If I use variables in dockerfile, it doesn't cache the stage, i.e everytime the developer will make a change and build the image to see that change reflected, they'll have to wait for 2-3 minutes, which can be annoying if you fixing a bug, what do you think?

nemesisdesign:
> I found a problem with this. If I use variables in dockerfile, it doesn't cache the stage, i.e everytime the developer will make a change and build the image to see that change reflected, they'll have to wait for 2-3 minutes, which can be annoying if you fixing a bug, what do you think?

@atb00ker without the possibility of installing variations of the official modules, one key goal of OpenWISP is not met: the possibility of reusing, extending and adapt the software to your own need.
I think this deserves to be moved into a separate issue and investigated further.

atb00ker:
> OpenWISP is not met: the possibility of reusing, extending and adapt the software to your own need.
I think this deserves to be moved into a separate issue and investigated further

Okay, I'll open a issue for it and get back to it asap! :smile:

PR #89, opened 2020-01-10
[WIP][tests] Moving tests to selenium

### LINKED ISSUE #53: [test] Use selenium for robust testing
We should use selenium for more robust testing of the functionality.

atb00ker:
Due to the fact that shell script tests are not 100% reliable, we want to move to selenium tests!

Closes #53


tests/config.json
```diff
@@ -0,0 +1,6 @@
+{
```
atb00ker: Since there are some configurations in tests that may change as per the requirement of the organization using it, I created a json file that is parsed in the testcase class.
But it feels like a overkill now, what do you think about it? 
Is this a good way to go about it or can you advice on another better way?

nemesisdesign: I think is fine!

tests/runtests.py
```diff
@@ -0,0 +1,337 @@
...
+        self.assertIn('Nodes', self.base_driver.page_source)
+
+
+    def test_topology_visual(self):
+
+        def _visualize_click():
+            # TODO: Ensure that we open the visualize page,
+            # I have tried waiting for the page to load but
+            # it doesn't work, for some reason I am not able
+            # to find, it doesn't work on the first click something.
```
atb00ker: I used this because hack because `visualize` page takes time to load and and the test fails randomly at times.
I tried a couple of tricks that didn't work:
1. `time.sleep()`
2. Using selenium [wait on element](https://stackoverflow.com/a/26567563/6410464)
3. wait for `document.ready`
4. explicitly checking if `visualize.js` is loaded before running tests.
5. using `DOMAIN/topology/topology/ID` page instead.

but none of the tricks worked reliably, I get the same result of race condition and test passing half of the time. 
Unfortunately, I am out of ideas.
Hence I am opening an issue for the same. :smile:

atb00ker: Removed this test from this PR, it already has another issue, added these details in that issue.

PR #93, opened 2020-01-21
[tests] Test wait_for_services and test_admin_login using selenium

marfgold1:
Help closes #89


marfgold1:
I've build it in my own fork and there's an error for `test_dashboard_login`. Seems like `/admin/` redirected to `/admin/login` causes the HTTP Response Code to be `302 Found` instead of `200 OK`.
You can see my build [here](https://travis-ci.org/marfgold1/docker-openwisp/builds/640065259)

README.md
```diff
@@ -181,18 +181,21 @@ Now you'll need to do steps (2) everytime you make a changes and want to build t
...
       username: username for logging in admin dashboard
       password: password for logging in admin dashboard
+      services_max_retries: Maximum number of retries to check if services are running
+      services_retry_wait: Delay time (in seconds) to each retries for checking if services are running
    ```
 
 5. Run tests:
 
    ```bash
+      python test_services.py
```
atb00ker: Instead of 2 different files, I would prefer that there is one file running both the tests! :smile:
Like `python runtests`

README.md
```diff
@@ -181,18 +181,21 @@ Now you'll need to do steps (2) everytime you make a changes and want to build t
       pip install selenium
    ```
 
-4. (Optional) Configure: open `tests/runtests.py` and configure variables as per your requirement, options are:
+4. (Optional) Configure: open `tests/config.json` and configure variables as per your requirement, options are:
```
atb00ker: Oopsie, my bad, thanks! :smile:

tests/test_admin.py
```diff
@@ -0,0 +1,41 @@
...
+import unittest
+from selenium import webdriver
+from selenium.webdriver.chrome.options import Options
+from selenium.common.exceptions import NoSuchElementException
+from utils import TestUtilities
+
+
+class TestAdmin(TestUtilities):
+
+    def setUp(self):
```
atb00ker: I think this is repeated, it can be moved to a parent class from where both the tests will get the common code! :smile:

tests/test_admin.py
```diff
@@ -0,0 +1,41 @@
...
+        config_file = os.path.join(os.path.dirname(__file__), "config.json")
+        with open(config_file) as json_file:
+            self.config = json.load(json_file)
+        self.chrome_options = Options()
+        self.chrome_options.add_argument('--ignore-certificate-errors')
+        if self.config['headless']:
+            self.chrome_options.add_argument("--headless")
+        self.base_driver = webdriver.Chrome(options=self.chrome_options)
+
+    def cleanUp(self):
```
atb00ker: Looks like `cleanup()` can go to the parent class as well! :smile:

tests/test_services.py
```diff
@@ -0,0 +1,44 @@
...
+import json
+import unittest
+from selenium import webdriver
+from selenium.webdriver.chrome.options import Options
+from utils import TestUtilities
+
+
+class TestServices(TestUtilities):
+
+    def setUp(self):
```
atb00ker: Again, DRY setUp() :smile:

tests/test_services.py
```diff
@@ -0,0 +1,44 @@
...
+        config_file = os.path.join(os.path.dirname(__file__), "config.json")
+        with open(config_file) as json_file:
+            self.config = json.load(json_file)
+        self.chrome_options = Options()
+        self.chrome_options.add_argument('--ignore-certificate-errors')
+        if self.config['headless']:
+            self.chrome_options.add_argument("--headless")
+        self.base_driver = webdriver.Chrome(options=self.chrome_options)
+
+    def cleanUp(self):
```
atb00ker: DRY cleanUp()

tests/test_websocket.py
```diff
@@ -52,15 +51,16 @@ def test_websocket(self):
...
+        self.log_in(self.change_location_driver, self.config['app_url'],
+                    self.config['username'], self.config['password'])
         self.get_location(self.change_location_driver,
                           self.config['app_url'], self.location_name)
         self.set_location_marker(self.change_location_driver,
                                  self.location_name)
         # assert that marker popped up on map
         self.assertEqual(len(self.base_driver
                                  .find_elements_by_class_name('leaflet-marker-icon')), 1)
+        print('SUCCESS: Websocket service is working!')
```
atb00ker: Print is not required. 
When this branch is merged to master the shell script will be removed and we will run tests like in other repositories! :smile:

tests/utils.py
```diff
@@ -3,6 +3,25 @@
 
 
 class TestUtilities(unittest.TestCase):
+
+    def check_dashboard(self, driver, app_url):
+        """
+        Check if admin dashboard is available
```
atb00ker: No, this is not required, I suppose a simple response of `200 OK` ensures the service is there, you don't need to do everything with selenium, this thing can even be done by a simply request: https://2.python-requests.org/en/master/

Moreover, this is only done once before starting tests, why is this in Utils, does this code need to be rerun? I can't think why I would rerun this! :sweat_smile:

tests/test_services.py
```diff
@@ -0,0 +1,44 @@
...
+        self.base_driver.close()
+
+    def test_services(self):
```
atb00ker: This can be put before the admin login test. Why does this have a seperate class.

For that fact, I just realized, why are you creating seperate class for each test, all the tests can be in one class, if there is a reason you are seperating them in files, please let me know! :smile:

atb00ker:
> I've build it in my own fork and there's an error for `test_dashboard_login`. 

Well, change the URL then, use the shell script only to get the idea what you have to do in the test, its not a good source for copying the code.

> Seems like /admin/ redirected to /admin/login causes the HTTP Response Code to be 302 Found instead of 200 OK

Well, I expect that, check the "user authentication" image here, it will tell you the order of requests: https://medium.com/@atb00ker/understanding-django-cookies-f018bc7a9092

tests/utils.py
marfgold1: That's exactly what I think, I've asked it in my gitter channel to use `requests` 😅

tests/test_admin.py
marfgold1: Should I do the same thing to websocket?

marfgold1:
@atb00ker Ah okay! I see the reason now, `test_dashboard_login` depends on the last login cookies (from `test_admin_login` I believe), and because I've moved the `test_admin_login` to selenium/python and remove the shell version of the test, `test_dashboard_login` does not have the login cookies yet :smile: 
The task said that I just have to move `wait_for_services` and `test_admin_login`, but consider GCI is almost done, should I also move `test_dashboard_login`? I'll merge `test_admin_login` and `test_dashboard_login` as one test case though because they were correlated.

marfgold1:
Okay, done! I make another class for `wait_for_services` because it's should be run first before running another task. For now, we can just test individual test case, and when we are done moving all the test, we can safely run `tests.py` just once :smile:

.travis.yml
```diff
@@ -19,7 +19,7 @@ before_script:
   - unzip chromedriver_linux64.zip -d /home/travis/bin
 
 install:
-  - pip install selenium
+  - pip install -r requirements-test.txt
```
atb00ker: Please just do `pip install selenium request`, requirements file is useful when this is getting complicated, right now, that's not the case.

Also, I forgot that `requests` is not a part of standard library, for something so simple, can please try to use the standard library? I think it's `urllib` (If it not that, ping me, I'll find the default one! :smile: )

tests/utils.py
```diff
@@ -118,3 +122,18 @@ def delete_location(self, driver, app_url, location_name):
         confirm_delete_button = \
             driver.find_element_by_xpath('//input[@type="submit"]')
         confirm_delete_button.click()
+
+
+class TestBase(TestUtilities):
```
atb00ker: Why is this not a part of TestUtilites?

marfgold1: Oh yeah I just realize that :sweat_smile:

marfgold1:
@atb00ker Wait... sorry I've just realize `.travis.yml` still in `pip install -r requirements.txt`. It's should be `pip install selenium`...
I'm so sorry I missed that one :cry:

atb00ker:
It's okay, It's not merged in master, I have to make some changes anyway, I'd wait for @nikitaermishin 's changes and then work on it before merging to master! :smile:

marfgold1:
Okay, thank you :smile:

Issue #91, opened 2020-01-20
[bug] admin namespace required while using API


atb00ker:
I just hit the error:
Controller version: `0.4`
URL: `http://controller.atb00ker-test.org/api/device-location/<device-uuid>/`
Exception Type: `NoReverseMatch`
Exception Value: `'admin' is not a registered namespace`

I am not sure if `admin` namespace should be required to use API, this might be a bug, I may need to investigate but opening a bug report for now.

PR #96, opened 2020-01-28
[bug] Fix APIs in containers #52 #91

### LINKED ISSUE #52: [improvement] remove APIs from dashboard container
Currently the dashboard container has APIs that belong in controller, radius and topology containers only.

- [ ] Update disable-services documentation.

atb00ker:
- Removed controller and topology API from dashboard
- Fixed controller API usage
- Moved OpenVPN to controller API

Closes #52 #91


build/openwisp_dashboard/urls.py
```diff
@@ -16,7 +15,5 @@
     url(r'^', include('openwisp_radius.urls', namespace='freeradius')),
 ]
 
-urlpatterns += openwisp_controller_urls()
-urlpatterns += openwisp_topology_urls()
```
nemesisdesign: great, are topology URLs served somewhere else right?

One quesiton, have you checked what happens when trying to visualize the topology graph in the admin?

atb00ker: @nemesisdesign 

Thanks for pointing this out, no it's not working, I am getting an error : `django.urls.exceptions.NoReverseMatch: Reverse for 'receive_topology' not found. 'receive_topology' is not a valid view function or pattern name.`

Either it needs to be added in admin namespace or topology `api` needs to stick with dashboard! :smile:

nemesisdesign: I think that we need to start to make the modules configurable about the base URL of the APIs that are called in the admin, there are quite several instances of this.

Alternatively for the moment yes we can also add the same views into the admin namespace if that does not generate too much duplication.

atb00ker:
Okay, now I am stuck with it.
I thought I could solve the problem with Monkey patching, so I wrote:
```
from openwisp_utils.admin import ReceiveUrlAdmin
from openwisp_network_topology.admin import TopologyAdmin


def receive_url(self, obj):
    """
    Monkey Patch `receive_url` in Topology Admin
    to avoid reverse URL lookup for `receive_url`.
    """
    url = "/api/receive/"
    if self.receive_url_querystring_arg:
        url = '{0}?{1}={2}'.format(
            url,
            self.receive_url_querystring_arg,
            getattr(obj, self.receive_url_querystring_arg)
        )
    return url


TopologyAdmin.receive_url = receive_url


class Media:
    """
    Monkey Patch Media to avoid loading the openwisp_utils's
    ReceiveUrlAdmin Media.
    """
    js = ('admin/js/jquery.init.js', 'receive_url.js')


ReceiveUrlAdmin.Media = Media
```

This works if I am using it inside `openwisp-network-topology/tests/admin.py`
But inside the docker repository, it fails. Infact,  even `self.receive_url_querystring_arg` is absent somehow in the repository. :frowning_face:

nemesisdesign:
I don't understand, why that monkey patching? We could fix that just by releasing django-netjsongraph and openwisp-network-topology, which is on our TODO list

nemesisdesign:
Oh right, now I understand. Well we should make that configurable. Better to invest in that rather than monkey patching, don't you think? Can you open an issue in openwisp-utils please?

atb00ker:
**Depends on : https://github.com/openwisp/openwisp-utils/pull/78**

build/openwisp_dashboard/urls.py
```diff
@@ -14,9 +18,27 @@
...
+        topology_url(urlpatterns, regex, receive_name)
+        break
+    elif api_url.url_patterns:
+        for path in api_url.url_patterns:
+            if hasattr(path, 'name') and path.name == receive_name:
+                path_regex = str(path.pattern)[1:] if str(path.pattern)[0] == '^' else str(path.pattern)
+                regex = str(api_url.pattern) + path_regex
+                topology_url(urlpatterns, regex, receive_name)
+                break
+
```
atb00ker: What do you think about this solution? 
I am copying the path from `topology_urls.py` file and registering it in like: 
```
    receive_url = url(pattern,
                      RedirectView.as_view(url=os.environ['TOPOLOGY_DOMAIN']),
                      name=receive_name)
```
make sure that dashboard gets the path while the url points to topology container. :smile:

build/openwisp_dashboard/urls.py
```diff
@@ -1,22 +1,46 @@
...
+# Create a url for topology receive pattern
+def topology_url(urlpatterns, pattern, receive_name):
+    receive_url = url(pattern,
+                      topology_redirect_view,
+                      name=receive_name)
+    urlpatterns.append(receive_url)
+
+
+for api_url in topology_urls:
+    if hasattr(api_url, 'name') and api_url.name == receive_name:
```
atb00ker: What do you think about this hacky solution? 
I am copying the path from `topology_urls.py` file and registering it as a redirect url to same path but topology domain:
```
def topology_redirect_view(request, pk):
    url = "{0}://{1}{2}".format(request.scheme,
                                os.environ["TOPOLOGY_DOMAIN"],
                                request.get_full_path())
```
make sure that dashboard gets the path while the url points to topology container. :smile:

atb00ker:
Upstream blocker(s) have been merged!

nemesisdesign:
:+1:

PR #101, opened 2020-02-13
[requirements] Move to channels2

atb00ker:


build/common/openwisp/settings.py
```diff
@@ -93,21 +86,20 @@
...
 
 CACHES = {
-    "default": {
-        "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": "redis://"+REDIS_HOST+":6379/1",
-        "OPTIONS": {
-            "CLIENT_CLASS": "django_redis.client.DefaultClient",
+    'default': {
+        'BACKEND': 'django_redis.cache.RedisCache',
+        'LOCATION': 'redis://'+REDIS_HOST+':6379/1',
```
nemesisdesign: why don't you use `format` which is more pythonic?

```python
'redis://{}:6379/1'.format(REDIS_HOST),
```

Are we using python 3.7 or python 3.6?
Because with python 3.7 we can use f strings which are even better:

```python
f'redis://{REDIS_HOST}:6379/1',
```

atb00ker: Okay. I'll move all the strings settings to f-string as soon as possible. :smile:

PR #106, opened 2020-02-22
Fixed MultipleObjects return #97

### LINKED ISSUE #97: [bug] openwisp_controller.pki.models.MultipleObjectsReturned
Recently, I've started seeing an issue with `load_init_script.py`:

```
dashboard_1   | Operations to perform:
dashboard_1   |   Apply all migrations: account, admin, auth, authtoken, config, connection, contenttypes, geo, openwisp_radius, openwisp_users, pki, reversion, sessions, sites, socialaccount, topology
dashboard_1   | Running migrations:
dashboard_1   |   No migrations to apply.
dashboard_1   | Traceback (most recent call last):
dashboard_1   |   File "load_init_data.py", line 139, in <module>
dashboard_1   |     os.environ['X509_NAME_CERT'])
dashboard_1   |   File "load_init_data.py", line 83, in create_default_cert
dashboard_1   |     return Cert.objects.get(name=x509NameCert)
dashboard_1   |   File "/usr/local/lib/python3.7/site-packages/django/db/models/manager.py", line 82, in manager_method
dashboard_1   |     return getattr(self.get_queryset(), name)(*args, **kwargs)
dashboard_1   |   File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 403, in get
dashboard_1   |     (self.model._meta.object_name, num)
dashboard_1   | openwisp_controller.pki.models.MultipleObjectsReturned: get() returned more than one Cert -- it returned 2!
```

It shouldn't happen because to avoid creating multiple Certs, `if not Cert.objects.filter(name=x509NameCert).exists():` was added. But still somethow the Cert is created. Need to investigate.

nepython:
Though the commit resolves the return of multiple objects, it doesn't solve  their creation in the first place.

Fixes #97


build/openwisp_dashboard/load_init_data.py
```diff
@@ -80,7 +80,7 @@ def create_default_cert(vpnOrg, defaultCa, x509NameCert):
         defaultCert.full_clean()
         defaultCert.save()
         return defaultCert
-    return Cert.objects.get(name=x509NameCert)
+    return Cert.objects.filter(name=x509NameCert).distinct()
```
atb00ker: I don't think this fixes the problem, it only applies a patch to hide the error. I don't think it's good for us to hide the error. I'd rather want to fix the core of the issue, 2 certificates shouldn't be created in the first place.
It only happens sometimes but it's a disaster when it happens really!

Regarding the travis failure: 
Please check the travis logs, I see: 
```
- missing prefix in the commit short description

  Eg: "[module/file/feature] Action performed"

Please read our guidelines at: http://openwisp.io/docs/developer/contributing.html#commit-message-style-guidelines

ERROR: Commit name check failed!
```

Issue #107, opened 2020-02-22
[improvements] Use f-string for all strings


atb00ker:
Check comment here: https://github.com/openwisp/docker-openwisp/pull/101#discussion_r382830175

nepython:
Hi!
Can I take up this issue?
I will do it asap.

jacobthebanana:
Hello! A quick `grep` search on `master` (as it is after commit [421589f](https://github.com/openwisp/docker-openwisp/commit/421589fe339bcd26e9faf014360ecede3ffcd65d) from two days ago) revealed that about half of the `format`'s and string concatenations in this repo are tests. Would you like to have these turned into f-strings as well? 

Here are the matches:
```bash
$ grep --regexp="[[\"\']\s*\+]*[.format\(]*" --extended-regexp --recursive --line-number docker-openwisp
```
```python
docker-openwisp/build/common/openwisp/settings.py:85:CELERY_BROKER_URL = 'redis://'+REDIS_HOST+':6379/1'
docker-openwisp/build/common/openwisp/settings.py:127:        'LOCATION': 'redis://'+REDIS_HOST+':6379/1',
docker-openwisp/build/common/services.py:32:                   ":" + os.environ['DASHBOARD_APP_PORT'])
docker-openwisp/build/common/utils.py:77:            parts = urlsplit('//' + addr)
docker-openwisp/build/common/utils.py:91:        return (host, '/' + url)
docker-openwisp/build/common/utils.py:149:        var['HTTP_' + key.strip().upper().replace('-', '_')] = value.strip()
docker-openwisp/build.py:18:                         r'DJANGO_SECRET_KEY=' + keygen,
docker-openwisp/build.py:23:        file_string += "DJANGO_SECRET_KEY=" + keygen
docker-openwisp/tests/runtests.py:114:                print("Unable to delete resource at: " +
docker-openwisp/tests/runtests.py:125:            print("One of the containers are down!\nOutput:\n" +
docker-openwisp/tests/runtests.py:263:            self.fail("Not all celery / celery-beat tasks are registered\nOutput:\n" +
docker-openwisp/tests/runtests.py:264:                      output + "\nError:\n" + error)
docker-openwisp/tests/runtests.py:280:            self.fail("Request not Accepted!\nOutput:\n" +
docker-openwisp/tests/runtests.py:281:                      output + "\nError:\n" + error)
[...]

PR #108, opened 2020-02-22
[improvements] Used f-string for all strings

hevalhazalkurt:
Changed the strings to the f-strings.

### DIFF
--- a/build.py
+++ b/build.py
@@ -15,12 +15,12 @@ def change_secret_key(keygen):
     file_string = file_handle.read()
     file_handle.close()
     file_string = re.sub(r'DJANGO_SECRET_KEY=.*',
-                         r'DJANGO_SECRET_KEY=' + keygen,
+                         r'DJANGO_SECRET_KEY={}'.format(keygen),
                          file_string)
     if file_string[-1] != "\n":
         file_string += "\n"
     if "DJANGO_SECRET_KEY" not in file_string:
-        file_string += "DJANGO_SECRET_KEY=" + keygen
+        file_string += "DJANGO_SECRET_KEY={}".format(keygen)
     file_handle = open(".env", 'w')
     file_handle.write(file_string)
     file_handle.close()
--- a/build/common/openwisp/settings.py
+++ b/build/common/openwisp/settings.py
@@ -68,7 +68,7 @@ WSGI_APPLICATION = 'openwisp.wsgi.application'
 ASGI_APPLICATION = 'openwisp_controller.geo.channels.routing.channel_routing'
 
 REDIS_HOST = os.environ['REDIS_HOST']
-CELERY_BROKER_URL = 'redis://'+REDIS_HOST+':6379/1'
+CELERY_BROKER_URL = f'redis://{REDIS_HOST}:6379/1'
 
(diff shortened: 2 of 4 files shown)


build.py
```diff
@@ -20,7 +20,7 @@ def change_secret_key(keygen):
     if file_string[-1] != "\n":
         file_string += "\n"
     if "DJANGO_SECRET_KEY" not in file_string:
-        file_string += "DJANGO_SECRET_KEY=" + keygen
+        file_string += f"DJANGO_SECRET_KEY={keygen}"
```
atb00ker: This runs outside the container and python version depends on the host machine.
Hence, can you please use the `format` function here please?  :smile:  
Thanks

atb00ker:
P.S: Please link the issue you are solving: https://github.com/openwisp/docker-openwisp/issues/107

hevalhazalkurt:
Sure. I've checked the project files but I could miss some strings. I'll check again and make the changes.

hevalhazalkurt:
Closes #108

build.py
```diff
@@ -20,7 +20,7 @@ def change_secret_key(keygen):
     if file_string[-1] != "\n":
         file_string += "\n"
     if "DJANGO_SECRET_KEY" not in file_string:
-        file_string += "DJANGO_SECRET_KEY=" + keygen
+        file_string += "DJANGO_SECRET_KEY={}".format(keygen)
```
nemesisdesign: this is not an f string, right?

hevalhazalkurt: @nemesisdesign yes that's right. this is the `.format()` function so it can also work on other Python versions such as Python 2.7 - 3.4. etc.

atb00ker: @nemesisdesign yes, the `build.py` runs during build on host machine and the user can have any python version! :smile:

atb00ker:
Also, there are more strings that need to be converted:
I see more strings in :
- build.py
- build/common/services.py
- build/common/utils.py
- build/common/openwisp/settings.py

Can you please double check these files? Thanks! :smile:

hevalhazalkurt:
Sorry, it's my first commit for somebody's repo so it's a little bit scary and unknown step for me. That's why I haven't touched some strings that I wasn't sure and didn't want to mess up your env. 😅Hope this one will be OK but it won't, I'll work again. 

Thank you for your understandings @atb00ker @nemesisdesign  😊

atb00ker:
That's not a problem. You can do `make runtests` to make sure everything is working as it should.
Secondly, now there are 3 commits, you want to rebase the commits.
```
git rebase -i <hash-of-commit-from-master>
# squash all your commits to one
git push -f
```
This will ensure there is only one commit. You can verify  by typing `git log` ! :smile:

build.py
nemesisdesign: ok, I'd add a comment then to make it clear, eg: `# no f-string because this is run on host machine, user may not have python 3.7`.

atb00ker:
Hi, I assume you have abandoned this task, have you?

atb00ker:
Closed by https://github.com/openwisp/docker-openwisp/pull/129

PR #111, opened 2020-03-10
[docker] Run containers as a non-root user

### LINKED ISSUE #109: [docker] run containers as a non-root user
It's not the best idea to run docker containers as root in production: https://americanexpress.io/do-not-run-dockerized-applications-as-root/

We should create user `openwisp` and run the containers as `openwisp` user with access to only the content it needs.

nepython:
Closes #109

### DIFF
--- a/build/openwisp_base/Dockerfile
+++ b/build/openwisp_base/Dockerfile
@@ -43,6 +43,9 @@ RUN pip install channels_redis service_identity django-redis psycopg2 uwsgi \
 ARG DJANGO_SOURCE=django==2.2
 RUN pip install --upgrade --prefix='/install' ${DJANGO_SOURCE}
 
+RUN useradd -u 8877 openwisp
+USER openwisp
+
 FROM SYSTEM
 
 COPY --from=PYTHON /install /usr/local
--- a/build/openwisp_freeradius/Dockerfile
+++ b/build/openwisp_freeradius/Dockerfile
@@ -3,6 +3,9 @@ FROM freeradius/freeradius-server:3.0.19-alpine
 RUN apk add --no-cache --update tzdata postgresql-dev postgresql-client && \
     rm -rf /var/cache/apk/* /tmp/*
 
+RUN adduser -u 8877 openwisp --disabled-password
+USER openwisp
+
 CMD ["sh", "init_command.sh"]
 EXPOSE 1812/udp 1813/udp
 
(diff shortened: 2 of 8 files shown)


nepython:
@atb00ker, not quite sure why build is showing, `NoSuchElementException: Message: no such element: Unable to locate element: {"method":"name","selector":"username"}` upon calling `driver.find_element_by_name('username').send_keys(username)`, I think it is occurring because page is not loading before it being called. I tried two changes for the same `self.login(username='openwisp')` `driver.implicitly_wait(15)` to solve this, but it still shows the same exception. Anything in this regard would be highly helpful as I am new to `selenium` it might be due to a very small mistake.:smiley:

atb00ker:
Hi,
Did you use `make develop-runtests` to test your changes in your work machine? Does it work there? :smile:
These are selenium tests getting that error means that some page (I think login page is not available.)

atb00ker:
>  Anything in this regard would be highly helpful as I am new to selenium it might be due to a very small mistake. :smiley:

Before you make sure that selenium tests pass, manually test if the page is even reachable.

nepython:
I did run `sudo make develop-runtests` (not quite sure why sudo was required to run this command on my local pc), it gave me the same error.
Also I had to add
```
chrome_options.add_argument('--no-sandbox')
chrome_options.add_argument('--disable-dev-shm-usage') 
```
Once again not known to me why I needed to add them but this was suggested in few forums as a work around for another error I was getting priorly.
> manually test if the page is even reachable.

Sure, will do this
:smiley:

atb00ker:
> not quite sure why sudo was required to run this command on my local pc

What fails when you don't use `sudo`? Is your docker setup done correctly? Can you run any other container without `sudo`?

```
chrome_options.add_argument('--no-sandbox')
chrome_options.add_argument('--disable-dev-shm-usage') 
```
This is because you are using `sudo`, `chromium / google chrome` are not designed to work as `root` and this is the "workaround".

Only once I have any details about the error, I will be able to tell why this is happening.

One useful thing for debugging:
Set headless to false here for debugging: https://github.com/openwisp/docker-openwisp/blob/master/tests/config.json

That will enable you to see what step is failing! :smile:

atb00ker:
Hi,

Closing this for now because no work has been done in an year, If you intend to continue, please reopen! :smile:

PR #133, opened 2020-10-13
Feature/build images on gitlab

atb00ker:


atb00ker:
After putting a lot of time into gitlab-ci,
I've had to install too many things in gitlab-ci to run the tests. The gitlab-ci requirements practically need another docker image of itself to avoid future version incompatibility problems. This will definitely add maintenance overhead.

So, I am considering using travis and pushing image from travis itself.

Any comments before I undo this image?

(P.S: I need to fix chromedriver version and find a solution for `the input device is not a TTY` error, so gitlab-ci not complete)

nemesisdesign:
@atb00ker using travis sounds good to me. We're already using it and we're familiar with it so we should stick to it.

atb00ker:
Temporary saving these changes in another branch until @hispanico gets back to this, meanwhile continuing with travis. :-)

PR #136, opened 2021-01-02
[update] Images openwisp version

atb00ker:


atb00ker:
Travis is failing because credentials are missing, pinged @nemesisdesign on gitter about it.
Merging after pushing images from local! :smile:

PR #137, opened 2021-01-02
[ux] Easy install script

atb00ker:
Script to easily install docker-openwisp for home users with basic configuration changes.

dockerOpenWISPEasy


nemesisdesign (commented):
This looks like a really great idea @atb00ker. 

Is there a way for asking the main domain? So that once the main domain is given, the subsequent defaults are based on the initial domain given, eg:

enter domain: <...>
user enters `mynetwork.net`

dashboard (dashboard.mynetwork.net): ...

atb00ker:
Can be done, will make this change soon! :smile:

atb00ker:
@nemesisdesign 

> Is there a way for asking the main domain? So that once the main domain is given, the subsequent defaults are based on the initial domain given, eg:

I looked into it, even if I apply this, it wouldn't reduce the number of questions because I'd still need to ask "Do you want to enable radius module"?

What do you think? Or should I enable these modules by default? :smile:

nemesisdesign:
> @nemesisdesign
> 
> > Is there a way for asking the main domain? So that once the main domain is given, the subsequent defaults are based on the initial domain given, eg:
> 
> I looked into it, even if I apply this, it wouldn't reduce the number of questions because I'd still need to ask "Do you want to enable radius module"?
> 
> What do you think? Or should I enable these modules by default?

I think the questions should be ask but a good default should be set so users can just press enter, we may even suggest users to do so (pressing enter continuously) if they want to do a quick test. Ideally they can just start the script, press enter continuously and get the openwisp instance running locally to play with it. Is this doable?

atb00ker:
homesExplain

Reduced input! :smile:

nemesisdesign (commented):
> homesExplain
> 
> Reduced input!

That's really nice @atb00ker!

Do you know why the build is failing?

atb00ker:
I fixed the build, added the docs, I'll check again tomorrow and merge it! :smile:

PR #146, opened 2021-02-27
[ux] Simplify .env

atb00ker:
Some users found the `.env` intimidating for trying out docker-openwisp, so I shaved the file mentioning the link for more settings! :smile:


.env
```diff
@@ -9,57 +12,16 @@ DB_USER=admin
 DB_PASS=admin
 # Security
 DJANGO_SECRET_KEY=default_secret_key
-DJANGO_ALLOWED_HOSTS=.openwisp.org
```
nemesisdesign, line 12: why is this being removed? Will it be written by the easy install script?

nemesisdesign (commented):
The simplification looks great!

.env
atb00ker: > why is this being removed?

I think people who don't know Django wouldn't simply get the concept, so I removed it.
By default `*` hosts are allowed.
I think any user who will take openwisp to production will see this option while customizing `docker-openwisp` for their organization, while testing though `*` should do just fine.

> Will it be written by the easy install script?

Yes. It's added by easy install script. (Made change after reading review)
`set_env "DJANGO_ALLOWED_HOSTS" ".$hostname"`

nemesisdesign: Ok, but I suggest to make sure this is mentioned somewhere, keep in mind [DJANGO_ALLOWED_HOSTS was added to prevent some type of attacks](https://docs.djangoproject.com/en/3.1/topics/security/#host-headers-virtual-hosting), so having the default set to `*` is less secure.
Can't you use the initialization script to automatically fill this setting based on the domain entered by the user?

atb00ker: I agree, in pragmatic programmer, there is mention that we should choose secure default values.
Hence, I have set: if no value is entered for `DJANGO_ALLOWED_HOSTS`, only dashboard's root domain is allowed. :smile:

PR #160, opened 2021-07-17
[config] Added new configuration options

atb00ker:


build/common/openwisp/settings.py
```diff
@@ -4,6 +4,17 @@
...
+for config in os.environ:
+    if 'OPENWISP_' in config:
+        value = os.environ[config]
+        if value.isdigit():
+            globals()[config] = int(value)
+        if value == 'None':
+            globals()[config] = None
+        else:
+            globals()[config] = value
+
```
atb00ker, line 17: Hi @nemesisdesign , what do you think about reading all the env vars and converting them into django settings variables.
Can you think of any security problem because of it?

nemesisdesign: I don't think it's a big issue, if someone can set the env var then can probably do other nasty things, what about boolean values or dictionaries though?

atb00ker: Added arrangements to handle bool, dict, the user will need to pass as json and then it'll be parsed.
Other more complex cases would need to be handled separately. :smile:

PR #161, opened 2021-07-17
[config] Removed default VPN organization

atb00ker:


nemesisdesign (commented):
having the default management VPN and VPN client template as shared is the most useful thing

PR #196, opened 2022-03-10
[ci] Updated chrome driver to version 99.0.4844.35

nemesisdesign:


.github/workflows/branch.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         run: |
           sudo apt update
           sudo apt install --yes google-chrome-stable
-          wget https://chromedriver.storage.googleapis.com/97.0.4692.20/chromedriver_linux64.zip
+          wget https://chromedriver.storage.googleapis.com/99.0.4844.35/chromedriver_linux64.zip
           sudo unzip chromedriver_linux64.zip -d /usr/local/bin/
           sudo chmod +x /usr/local/bin/chromedriver
```
pandafy, line 37: chromedriver comes pre-installed in the workflow image file. It is not required to install it.

atb00ker: That's possible, these github-action files are old! 
Let me test them without trying to install chromium! 😄

PR #198, opened 2022-03-14
[change] Enable gzip compression

nemesisdesign:
Enable gzip compression by default.


nemesisdesign (commented):
I tried double checking whether the gzip_static nginx module is working using these tips:
https://stackoverflow.com/questions/2460821/how-can-i-check-that-the-nginx-gzip-static-module-is-working, but in the current nginx container it seems it's not possible to use strace due to the restrictive permissions.

atb00ker (approved):
LGTM! 😄

PR #199, opened 2022-03-14
[change] Improved build of openwisp_base python deps

nemesisdesign:
The build is taking an excessively long time to be executed, it looks to me that a lot of unnecessary actions are being performed so I spent some time cleaning up things that do not look ok to me.

- Removed obsolete django-rest-auth (we use a maintained fork called [dj_rest_auth](https://github.com/iMerica/dj-rest-auth) defined in OpenWISP RADIUS, so there's no need to redefine it here)
- Define additional requirements in requirements.txt, this will also allow us to monitor dependencies more easily
- Do not attempt installing custom openwisp modules unless a non default argument is supplied: there's no need to install again openwisp-users, openwisp-utils, django-x509 and even openwisp-controller unless a non default argument is supplied, because these modules are installed automatically in the previous pip install commands
- Avoid reinstalling every time to save build time - this was causing all the dependencies to be uninstalled and reinstalled each time, surely a waste of energy and time
- No need to use pip==20.2.4 anymore

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,18 +133,17 @@ You can build with your own python package by creating a file named `.build.env`
 - `OPENWISP_NOTIFICATION_SOURCE`
 - `OPENWISP_TOPOLOGY_SOURCE`
 - `OPENWISP_RADIUS_SOURCE`
+- `OPENWISP_IPAM_SOURCE`
 - `OPENWISP_USERS_SOURCE`
 - `OPENWISP_UTILS_SOURCE`
-- `OPENWISP_IPAM_SOURCE`
-- `DJANGO_SOURCE`
-- `DJANGO_NETJSONCONFIG_SOURCE`
 - `DJANGO_X509_SOURCE`
+- `DJANGO_SOURCE`
 
 For example, if you want to supply your own django and openwisp-controller source, your `.build.env` should be written like this:
 
 ```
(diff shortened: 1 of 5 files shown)


nemesisdesign:
@atb00ker comparing the build times of this PR and the other one for the gzip enabling, take a look:

[[change] Improved build of openwisp_base python deps](https://github.com/openwisp/docker-openwisp/runs/5544231500) took 4 minutes and 6 seconds to build.
[[change] Enable gzip compression](https://github.com/openwisp/docker-openwisp/runs/5544099868) took 18 minutes and 36 seconds.

Quite a difference!

atb00ker (approved):
The changes look good, the appox build time is ~10 minutes.
We should re-run this build, if we don't come near ~10 minute mark, then will need to investigate.
(removing `--force-install` might be potential cause, in `pip20.04+` algorithm, it wastes time trying to find the correct version [or at least it used to do that] )

nemesisdesign:
> The changes look good, the appox build time is ~10 minutes. We should re-run this build, if we don't come near ~10 minute mark, then will need to investigate. (removing `--force-install` might be potential cause, in `pip20.04+` algorithm, it wastes time trying to find the correct version [or at least it used to do that] )

The force reinstall was taking a huge time on my machine.

PR #200, opened 2022-03-15
[change] Create dedicated openwisp user #109

### LINKED ISSUE #109: [docker] run containers as a non-root user

nemesisdesign:
Work in progress, not ready yet, but functional.

Will close #109

### DIFF
--- a/build/common/utils.sh
+++ b/build/common/utils.sh
@@ -1,7 +1,6 @@
 #!/bin/sh
 
 function init_conf {
-    ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
     default_psql_vars
 }
 
--- a/build/openwisp_base/Dockerfile
+++ b/build/openwisp_base/Dockerfile
@@ -4,62 +4,79 @@ FROM python:3.10-slim-bullseye AS SYSTEM
 # 1. gettext: Required by envsubst used in scripts
 # 2. openwisp-radius/weasyprint: libcairo2 libpangocairo-1.0
 # 3. openwisp-monitoring: fping gdal-bin
+# 4. sudo
 RUN apt update && \
-    apt install --yes --no-install-recommends libcairo2 libpangocairo-1.0-0 gdal-bin \
-                                              gettext fping \
-    && rm -rf /var/lib/apt/lists/* /root/.cache/pip/* /tmp/*
+    apt install --yes --no-install-recommends \
+                libcairo2 apt-utils libpangocairo-1.0-0 \
+                gdal-bin gettext fping sudo && \
+    rm -rf /var/lib/apt/lists/* /root/.cache/pip/* /tmp/*
+
+RUN apt update && \
+    apt install --yes --no-install-recommends \
+                gcc libpq-dev libjpeg-dev libffi-dev python3-dev \
+                python3-pip libxml2-dev libxslt1-dev zlib1g-dev g++
+
+RUN useradd --system --password '' --create-home --shell /bin/bash \
+   --gid root --groups sudo --uid 1001 openwisp
+RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
+USER openwisp:root
 
 FROM SYSTEM AS PYTHON
 
-RUN apt update && \
-    apt install --yes --no-install-recommends gcc libpq-dev libjpeg-dev libffi-dev python3-dev \
-                                              python3-pip libxml2-dev libxslt1-dev zlib1g-dev g++
+ENV PATH="${PATH}:/home/openwisp/.local/bin"
+ENV PYTHONPATH=/home/openwisp/.local/lib/python3.10/site-packages
 
-ENV PYTHONPATH=/install/lib/python3.10/site-packages
+RUN pip install --user --upgrade pip
 ARG OPENWISP_MONITORING_SOURCE=https://github.com/openwisp/openwisp-monitoring/tarball/docker-openwisp-usage
-RUN pip install --prefix='/install' --upgrade ${OPENWISP_MONITORING_SOURCE}
+RUN pip install --user --upgrade ${OPENWISP_MONITORING_SOURCE}
 ARG OPENWISP_FIRMWARE_SOURCE=https://github.com/openwisp/openwisp-firmware-upgrader/tarball/docker-openwisp-usage
-RUN pip install --prefix='/install' --upgrade ${OPENWISP_FIRMWARE_SOURCE}
+RUN pip install --user --upgrade ${OPENWISP_FIRMWARE_SOURCE}
 ARG OPENWISP_TOPOLOGY_SOURCE=https://github.com/openwisp/openwisp-network-topology/tarball/docker-openwisp-usage
-RUN pip install --prefix='/install' --upgrade ${OPENWISP_TOPOLOGY_SOURCE}
+RUN pip install --user --upgrade ${OPENWISP_TOPOLOGY_SOURCE}
 ARG OPENWISP_RADIUS_SOURCE=https://github.com/openwisp/openwisp-radius/tarball/docker-openwisp-usage
-RUN pip install --prefix='/install' --upgrade ${OPENWISP_RADIUS_SOURCE}
+RUN pip install --user --upgrade ${OPENWISP_RADIUS_SOURCE}
 ARG OPENWISP_IPAM_SOURCE=https://github.com/openwisp/openwisp-ipam/tarball/docker-openwisp-usage
-RUN pip install --prefix='/install' --upgrade ${OPENWISP_IPAM_SOURCE}
+RUN pip install --user --upgrade ${OPENWISP_IPAM_SOURCE}
 
 # here we try to install custom versions of the modules only if the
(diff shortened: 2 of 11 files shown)


build/common/utils.sh
```diff
@@ -1,7 +1,6 @@
 #!/bin/sh
 
 function init_conf {
-    ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
```
nemesisdesign, line 4: @atb00ker is this really needed? I have noticed that also without this the timezone is set correctly in the containers (checked using the `date` command).

atb00ker: The logs might display time in UTC without this command because, I think supervisord relies on the file in `/etc/localtime` to get the timezone for logs.

But if all the logs show correct time without this command now, I am good with removing it! 😄

build/openwisp_base/Dockerfile
```diff
@@ -1,65 +1,77 @@
...
+RUN pip install --user --upgrade -r /tmp/openwisp-deploy-requirements.txt
 ARG DJANGO_SOURCE=django~=4.0.0
-RUN pip install --prefix='/install' --upgrade ${DJANGO_SOURCE}
+RUN pip install --user --upgrade ${DJANGO_SOURCE}
 
 FROM SYSTEM
 
-COPY --from=PYTHON /install /usr/local
-COPY ./common/ /opt/openwisp/
+COPY --from=PYTHON --chown=openwisp:root /home/openwisp/.local/ /usr/local
```
atb00ker, line 72: Might be worth checking if it's increasing the size of the container by a lot, but otherwise, GG! 😄

atb00ker (approved):
Looks good! 😄

atb00ker (commented):
The following should be checked before merging this:
- Uploading CSV
- Uploading floorplans
- Downloading PDF
- Connecting to a device with SSH method and checking if an update is pushed properly without permission issues.


and, the rest looks good! 😄

nemesisdesign (commented):
I still need to check some other containers which I believe are still running as root.

nemesisdesign (commented):
Status:

- openwisp application containers are running as user `openwisp`
- openvpn starts as root and drops down to `nobody:nogroup` after initialization
- nginx is already set to run with the `nginx` user
- freeradius is running as root and I haven't been able to change this yet

With freeradius I tried this:

```diff
diff --git a/build/openwisp_freeradius/Dockerfile b/build/openwisp_freeradius/Dockerfile
index dd285bd..aaf1224 100644
--- a/build/openwisp_freeradius/Dockerfile
+++ b/build/openwisp_freeradius/Dockerfile
@@ -1,14 +1,21 @@
 FROM freeradius/freeradius-server:3.0.25-alpine
 
-RUN apk add --no-cache --update tzdata postgresql-dev postgresql-client && \
+RUN apk add --no-cache --update tzdata postgresql-dev postgresql-client sudo && \
     rm -rf /var/cache/apk/* /tmp/*
 
+RUN addgroup -S freerad && \
+    addgroup -S sudo && \
+    adduser -S freerad -G freerad -G sudo
+
+RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
+USER freerad
+
 CMD ["sh", "init_command.sh"]
 EXPOSE 1812/udp 1813/udp
 
-COPY ./common/init_command.sh \
+COPY --chown=freerad:freerad ./common/init_command.sh \
     ./common/utils.sh ./
-COPY --chown=root:root ./openwisp_freeradius/raddb/ /etc/raddb/
+COPY --chown=freerad:freerad ./openwisp_freeradius/raddb/ /etc/raddb/
 
 ENV TZ=UTC \
     MODULE_NAME=freeradius \
```

But after changing the user with `USER freerad` the radiusd command is not called.

BTW I am trying to run the image without those changes but even then I cannot find `radiusd` on the filesystem, I tried to look at the source code of the base image to find out what kind of magic they're doing but I can't find the source code.

nemesisdesign (commented):
@atb00ker I am really having a hard time fixing the uwsgi build issue.

I tried installing uwsgi via package manager as an alternative but I see it complains about several other things, including not being able to find the installed python packages, so there's quite a few things that I am not understanding..:

- why does the build fail on github actions but succeed for me locally? Is there anything I can do to make the local build look more what happens on github actions?
- I think I am not understanding how the `FROM` statements are used in this `Dockerfile`, when can I study that kind of usage? I have an impression it has to do with the build errors, maybe when pip tries to install uwsgi it's missing some dependency that it's installed in the previous later and not available in the subsequent layer?

nemesisdesign:
This worked, although it's ugly: https://github.com/openwisp/docker-openwisp/pull/200/commits/feb29887969678ad2b56fb4244e28ba654162181. I really don't understand why that would work and what came before didn't.

I am going to try another variant shortly.

nemesisdesign:
No idea how the latest changes worked while the previous ones didn't because uwsgi couldn't be built!
If anyone knows please explain this to me.. does it have to do anything with build stages? Maybe gcc was not available to pip when having to compile uwsgi? But in that case why now it does work?! No idea and it bothers me..

atb00ker:
CC from Gitter chat: 

We are using multi stage Dockerfiles.

After FROM SYSTEM everything available in steps above is not available, except of the packages installed in /install directory, which is copied from the previous image.

The reason for this is 2 fold.

    Size: the size can more than double in normal docker build.
    Security: we don't want to make development deps available in prod (like GCC)

build/openwisp_base/Dockerfile
nemesisdesign: These are the current sizes:

```
CONTAINER ID   IMAGE                                 COMMAND                  CREATED          STATUS          PORTS                                                                      NAMES                          SIZE
ceeddce1e1a9   openwisp/openwisp-dashboard:latest    "bash init_command.sh"   17 minutes ago   Up 17 minutes   8000/tcp                                                                   docker-openwisp_celery_1       35.9kB (virtual 994MB)
bbbdee687a12   openwisp/openwisp-api:latest          "bash init_command.sh"   17 minutes ago   Up 17 minutes   8001/tcp                                                                   docker-openwisp_api_1          33.9kB (virtual 994MB)
82af8c5f55e6   openwisp/openwisp-dashboard:latest    "bash init_command.sh"   17 minutes ago   Up 17 minutes   8000/tcp                                                                   docker-openwisp_celerybeat_1   52.3kB (virtual 994MB)
84bb4f818caf   openwisp/openwisp-websocket:latest    "supervisord --nodae…"   17 minutes ago   Up 17 minutes   8002/tcp                                                                   docker-openwisp_websocket_1    12.9kB (virtual 994MB)
3ca80bba5f70   openwisp/openwisp-dashboard:latest    "bash init_command.sh"   18 minutes ago   Up 17 minutes   8000/tcp                                                                   docker-openwisp_dashboard_1    38.2kB (virtual 994MB)
45042accb8d1   openwisp/openwisp-freeradius:latest   "/docker-entrypoint.…"   29 minutes ago   Up 29 minutes   0.0.0.0:1812-1813->1812-1813/udp, :::1812-1813->1812-1813/udp              docker-openwisp_freeradius_1   2.81kB (virtual 317MB)
7187641daf6a   openwisp/openwisp-nginx:latest        "/docker-entrypoint.…"   54 minutes ago   Up 54 minutes   0.0.0.0:80->80/tcp, :::80->80/tcp, 0.0.0.0:443->443/tcp, :::443->443/tcp   docker-openwisp_nginx_1        6.98kB (virtual 101MB)
cf8c5e6b576a   openwisp/openwisp-openvpn:latest      "sh init_command.sh"     54 minutes ago   Up 52 minutes   1194/tcp, 0.0.0.0:1194->1194/udp, :::1194->1194/udp                        docker-openwisp_openvpn_1      45.7kB (virtual 70.5MB)
886fc2e83c40   redis:alpine                          "docker-entrypoint.s…"   54 minutes ago   Up 54 minutes   6379/tcp                                                                   docker-openwisp_redis_1        0B (virtual 32.4MB)
01d543620f10   mdillon/postgis:11-alpine             "docker-entrypoint.s…"   54 minutes ago   Up 54 minutes   5432/tcp                                                                   docker-openwisp_postgres_1     63B (virtual 146MB)
0d5784cee9fe   influxdb:1.8-alpine                   "/entrypoint.sh infl…"   54 minutes ago   Up 54 minutes   8086/tcp                                                                   docker-openwisp_influxdb_1     0B (virtual 163MB)
803080d3fad2   openwisp/openwisp-postfix:latest      "sh init_command.sh"     54 minutes ago   Up 54 minutes   25/tcp                                                                     docker-openwisp_postfix_1      38.1kB (virtual 56.4MB)
```

build/common/utils.sh
nemesisdesign: The date command returns the right date in every container. Is there anything else I should try? That should be enough, if the date command returns the right date all the rest should follow since that's the system time.

nemesisdesign (commented):
It's finally ready now.

PR #202, opened 2022-03-18
[change] Added django.contrib.humanize

nemesisdesign:
Needed for https://github.com/openwisp/openwisp-radius/pull/394 .

### DIFF
--- a/build/openwisp_dashboard/module_settings.py
+++ b/build/openwisp_dashboard/module_settings.py
@@ -8,6 +8,7 @@ INSTALLED_APPS = [
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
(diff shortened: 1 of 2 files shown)


build/openwisp_api/module_settings.py
```diff
@@ -4,6 +4,7 @@
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
+    'django.contrib.humanize',
```
atb00ker, line 7: If it's required for radius, do we need it in API container? 🤔

nemesisdesign: I removed it, I think it's not needed, it's needed in the dashboard settings because that container is used for celery, the email is sent in the background.

PR #205, opened 2022-03-28
[Feature] Nginx security improvements #194

### LINKED ISSUE #194: [feature] Nginx security improvements
Port the nginx security improvements done recently in ansible-openwisp2 (disable SHA1 ciphers, remove old SSL/TLS versions, configure the Content-Security-Policy header).

codesankalp:
- Added csp header
- Removed old SSL/TLS versions
- Disabled SHA1 ciphers

Closes #194


codesankalp:
Console shows error related to content security policy:

```
https://dashboard.openwisp.org/static/admin/js/vendor/jquery/jquery.min.8fb8fee4fcc3.js 1 Refused to connect to 'https://api.openwisp.org/api/v1/monitoring/geojson/' because it violates the following Content Security Policy directive: "connect-src https://dashboard.openwisp.org wss: 'self'".

https://dashboard.openwisp.org/static/admin/js/vendor/jquery/jquery.min.8fb8fee4fcc3.js 1 Refused to connect to 'https://api.openwisp.org/api/v1/monitoring/geojson/' because it violates the following Content Security Policy directive: "connect-src https://dashboard.openwisp.org wss: 'self'".

https://dashboard.openwisp.org/static/admin/js/vendor/jquery/jquery.min.8fb8fee4fcc3.js 1 Refused to connect to 'https://api.openwisp.org/api/v1/monitoring/geojson/' because it violates the following Content Security Policy directive: "connect-src https://dashboard.openwisp.org wss: 'self'".

https://dashboard.openwisp.org/static/admin/js/vendor/jquery/jquery.min.8fb8fee4fcc3.js 1 Refused to connect to 'https://api.openwisp.org/api/v1/monitoring/geojson/' because it violates the following Content Security Policy directive: "connect-src https://dashboard.openwisp.org wss: 'self'".
```

atb00ker:
@codesankalp , please try going to the `api.openwisp.org` domain and ensure you can make calls to the API (any working endpoint will do), if it's working, then it might be worth investigating why it's happening, I have not seen it!

atb00ker:
P.S: If the problem is indeed because of the new headers, `*.<domain>` sounds good! 😄

build/common/utils.sh
```diff
@@ -103,13 +103,18 @@ function ssl_http_behaviour {
     fi
 }
 
+function domain {
+    eval "python3 -c \"import os;print('.'.join(os.environ['API_DOMAIN'].split('.')[-2:]), end='')\""
```
codesankalp, line 107:
```bash
#!/bin/bash

domain=$1
md=""
p2='^(.*\.)?([^.]+\.[a-z]+\.[a-z][a-z])$'
p3='^(.*\.)?([^.]+\.(com|org|net|int|edu|gov|mil))$'
px='^(.*\.)([a-z]+)$'
# 2 letter country codes
if [[ "$domain" =~ $p2 ]] ; then
    md=${BASH_REMATCH[2]};
# 3 letters legacy domain
elif [[ "$domain" =~ $p3 ]] ; then
    md=${BASH_REMATCH[2]};
# All Other
elif [[ "$domain" =~ $px ]] ; then
    md=${BASH_REMATCH[2]};
fi ;
echo "$md"

```

I tried this script yesterday, but it is not working in shell script. So i used python to get the domain.

atb00ker: Right now, I am using:
```
"import os;print('.'.join(os.environ['API_DOMAIN'].split('.')[1:]),
```

codesankalp: But shouldn't this thing fail for multi level sub domains like `api.web.openwisp.org` ?

codesankalp: Also my code will also fail for `api.openwisp.co.in`.

pandafy: If we are using python and we don't care about adding another dependency, we can use [tldextract](https://pypi.org/project/tldextract/) library which is built for this purpose.

build/openwisp_nginx/openwisp.ssl.template.conf
```diff
@@ -14,13 +14,22 @@ server {
...
     # Aditional Security Headers
-    add_header Strict-Transport-Security "max-age=31536000";
-    add_header X-Content-Type-Options nosniff;
+    add_header X-XSS-Protection          "1; mode=block" always;
+    add_header X-Content-Type-Options    "nosniff" always;
+    add_header Referrer-Policy           "same-site" always;
+    add_header Content-Security-Policy   "default-src http: https: data: blob: 'unsafe-inline';
+    script-src 'unsafe-eval' https: 'unsafe-inline' 'self';
+    frame-ancestors 'self'; connect-src https://${DOMAIN} wss: 'self';
+    worker-src https://${DOMAIN} blob: 'self';" always;
```
pandafy, line 30: Won't we need to make `Content-Security-Policy` configurable like it is in ansible-openwisp2?

pandafy (commented):
The nginx configuration looks good to me.

build/common/utils.sh
atb00ker: `tldextract` is also relying on an database to do this thing.
As of `api.web.openwisp.org`, yes, it will fail, but that's mentioned in the `ENV.md` file that you can't use a multipart subdomain.
As of `api.openwisp.co.in`, no it will not fail, please test it out, someone I know is using such a domain so, to the best of my knowledge it should work! 😄

requirements-test.txt
```diff
@@ -1,3 +1,3 @@
 docker~=5.0.3
 selenium~=4.1.0
-openwisp-utils[qa]~=1.0.0
+openwisp-utils[qa] @ https://github.com/openwisp/openwisp-utils/tarball/master
```
atb00ker, line 3: Why is [qa] in this file.
We should not be needing qa in the first place, looks like a recent change! :-O

nemesisdesign (commented):
> Why is [qa] in this file.
> We should not be needing qa in the first place, looks like a recent change! :-O
>
> (Might not be relevant to this PR at all. just noticed)

@atb00ker openwisp-utils qa is used here:
https://github.com/openwisp/docker-openwisp/blob/master/.github/workflows/branch.yml#L29
BTW we should have a run-qa-checks script and add more QA checks and style enforcements here if we want to reach the quality of the other repos, so I created  https://github.com/openwisp/docker-openwisp/issues/207.

build/openwisp_nginx/openwisp.ssl.template.conf
nemesisdesign: @codesankalp @atb00ker can we make it consistent with [how it's done in ansible-openwisp2](https://github.com/openwisp/ansible-openwisp2/pull/342/files)?
We could define this as a customizable variable in the docker image for example?

build/common/utils.sh
nemesisdesign: I think we should lean in favour of solutions like https://pypi.org/project/tldextract/ instead of implementing our own unless we can do it better in a simpler way.

In this case it looks like the people working on tldextract have done a great job and the library seems well maintained so definitely we should try it so hopefully we can get around the current limitations.

@codesankalp can you give it a try please?
Also try to have a dedicated script for dealing with this unless is a trivial line of code.

requirements-test.txt
nemesisdesign: Please @codesankalp when you need to fix failing builds on master, send a separate PR, I need this fix for the other PR too and I am pushing it to master shortly. It's duplicated work from different people.

nemesisdesign: Please remove this and rebase on master: https://github.com/openwisp/docker-openwisp/commit/9acc2001700cb9347d66ada65dedb5564901b3ab.

build/common/init_command.sh
```diff
@@ -41,6 +41,7 @@ elif [ "$MODULE_NAME" = 'openvpn' ]; then
     supervisord --nodaemon --configuration supervisord.conf
 elif [ "$MODULE_NAME" = 'nginx' ]; then
     rm -rf /etc/nginx/conf.d/default.conf
+    pip install tldextract
```
nemesisdesign, line 44: let's install this in the Dockerfile of the base image please, add it to the requirements.txt and pin the version.

build/openwisp_nginx/get_domain.py
```diff
@@ -0,0 +1,6 @@
+import sys
+import os
+import tldextract
+
+ext = tldextract.extract(os.environ['API_DOMAIN'])
+sys.stdout.write(ext.registered_domain)
```
nemesisdesign, line 6: nice and clean

build/openwisp_nginx/openwisp.ssl.template.conf
```diff
@@ -14,13 +14,22 @@ server {
...
     # Aditional Security Headers
-    add_header Strict-Transport-Security "max-age=31536000";
-    add_header X-Content-Type-Options nosniff;
+    add_header X-XSS-Protection          "1; mode=block" always;
+    add_header X-Content-Type-Options    "nosniff" always;
+    add_header Referrer-Policy           "same-site" always;
+    add_header Content-Security-Policy   "default-src http: https: data: blob: 'unsafe-inline';
+    script-src 'unsafe-eval' https: 'unsafe-inline' 'self';
+    frame-ancestors 'self'; connect-src *.${ROOT_DOMAIN} wss: 'self';
+    worker-src https://${DOMAIN} blob: 'self';" always;
```
nemesisdesign, line 30: Since this is a string, shoulnd't we be able to have the contents of this as an env variable and hence allow to customize it if needed?

eg:

```
add_header Content-Security-Policy ${NGINX_CONTENT_SECURITY_POLICY}
```

```
ENV MODULE_NAME=nginx \
# ....
    NGINX_CONTENT_SECURITY_POLICY="\"default-src http: https: data: blob: 'unsafe-inline'; script-src 'unsafe-eval' https: 'unsafe-inline' 'self'; frame-ancestors 'self'; connect-src *.${ROOT_DOMAIN} wss: 'self'; worker-src https://${DOMAIN} blob: 'self';\" always";
```
That would make it consistent with [how it's done in ansible-openwisp2](https://github.com/openwisp/ansible-openwisp2/pull/342/files).

nemesisdesign (changes requested):

build/openwisp_nginx/openwisp.ssl.template.conf
codesankalp: I added this but, I am getting this error:

image

nemesisdesign (commented):
Ok I think we can keep it simple for now, seems making the `Content-Security-Policy` overridable makes the code a lot more complicated, I think that should be overridable with `NGINX_SSL_CONFIG`, in future iterations we'll improve the override capability according to the usage we'll make of this.

nemesisdesign (approved):

PR #212, opened 2022-04-07
[feature] Added functionality to tune celery workers #191

### LINKED ISSUE #191: [feature] Allow to tune the celery concurrency scaling settings
Allow to tune the celery concurrency scaling settings, docker-openwisp does not allow this right now.

pandafy:
Closes #191


.env
```diff
@@ -41,6 +41,14 @@ VPN_CLIENT_NAME=default-management-vpn
 DEBUG_MODE=False
 DJANGO_LOG_LEVEL=INFO
 USE_OPENWISP_CELERY_TASK_ROUTES_DEFAULTS=True
+OPENWISP_CELERY_CONCURRENCY=1
```
nemesisdesign, line 44: what if we do something like:

```
OPENWISP_CELERY_COMMAND_FLAGS="--concurrency 1"
```

So that we can configure any flag more easily and avoid all those ifs below which would be quite painful to maintain?

This solution also allow for great flexibility in configuring the celery workers.

build/common/init_command.sh
```diff
@@ -82,11 +85,13 @@ elif [ "$MODULE_NAME" = 'celery_monitoring' ]; then
...
                -n monitoring@%h --logfile /opt/openwisp/logs/celery_monitoring.log \
-               --pidfile /opt/openwisp/celery_monitoring.pid --detach
+               --pidfile /opt/openwisp/celery_monitoring.pid --detach \
+               ${OPENWISP_CELERY_MONITORING_COMMAND_FLAGS}
         echo "Starting the 'monitoring_checks' celery worker"
         celery -A openwisp worker -l ${DJANGO_LOG_LEVEL} --queues monitoring_checks \
                -n monitoring_checks@%h --logfile /opt/openwisp/logs/celery_monitoring_checks.log \
-               --pidfile /opt/openwisp/celery_monitoring_checks.pid --detach
+               --pidfile /opt/openwisp/celery_monitoring_checks.pid --detach \
+               ${OPENWISP_CELERY_MONITORING_COMMAND_FLAGS}
```
nemesisdesign, line 94: I think here we should have a dedicated variable: `OPENWISP_CELERY_MONITORING_CHECKS_COMMAND_FLAGS`, so we can control the amount of workers for both queues, it's not 100% sure they have to be always the same.

nemesisdesign (changes requested):
Looks good, see my comment below.

nemesisdesign (commented):
Looks good, will test it.

nemesisdesign (approved):

PR #213, opened 2022-04-15
[qa] Added ./run-qa-checks script #207

### LINKED ISSUE #207: [qa] Add ./run-qa-checks script
Let's add a `./run-qa-checks` script as in the other repositories and let's call that from the CI build.

I think we should definitely auto-format the python and shell code as we're doing in the python modules and openwrt packages. Is there any good and popular linter for `Dockerfile` definitions? If yes it will be good to add that too.

codesankalp:
Closes #207


run-qa-checks
```diff
@@ -0,0 +1,34 @@
...
+echo 'Python file QA checks...'
+openwisp-qa-check --skip-checkmigrations
+
+echo ''
+echo 'Dockerfile QA checks...'
+
+# ignoring some dockerfile-lint:
+# DL3013 - Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>`
+# DL3007 - Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag
+# DL3018 - Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`
```
codesankalp, line 25: Ignoring these, to install current master and development version.

nemesisdesign (changes requested):
What about python code? There's also python code here and it would be great to format and check that as well so we avoid wasting time formatting it manually.
I see that black ignores by default anything which is contained in the `build` directory, see https://github.com/psf/black/issues/1193 (it makes sense, because when building python packages a build dir is created which contains a lot of extra files).

There are ways to work around this but it would be a lot simpler if we just renamed build to something else. Can we rename `build` to `images` @atb00ker?

atb00ker:
@nemesisdesign yes, renaming should not cause any problems or compatibility issues! 😄

.github/workflows/pull_request.yml
```diff
@@ -17,10 +17,12 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Install Testing Requirements
-        run: sudo pip install -r requirements-test.txt
+        run: |
+          sudo pip install -r requirements-test.txt
+          sudo curl -sL -o /bin/hadolint "https://github.com/hadolint/hadolint/releases/download/v1.16.3/hadolint-Linux-x86_64" && sudo chmod +x /bin/hadolint
```
atb00ker, line 22: \imo: The two commands can be in different lines for readability.

atb00ker (approved):

.github/workflows/pull_request.yml
codesankalp: I didn't separated the two command because it is used for installing one dependency `hadolint`. I will change this for readability purpose.

run-qa-checks
```diff
@@ -23,12 +23,12 @@ echo 'Dockerfile QA checks...'
 # DL3013 - Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>`
 # DL3007 - Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag
 # DL3018 - Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`
```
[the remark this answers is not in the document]
pandafy: I think this step will be right direction.

We will need to follow the release of dependency packages more closely. We can leverage "Depandabot" for that. 

On a personal note, I am fed up of random failing builds.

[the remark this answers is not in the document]
codesankalp: I agree in pinning the version and tagging build.
I have one doubt in python packages i.e. for openwisp packages we are using the github master tarball and we cannot specify version in this case, so should I ignore these lines?
https://github.com/openwisp/docker-openwisp/blob/ad61dfd26584d69de3651cd8935d2f0b73115740/images/openwisp_base/Dockerfile#L28-L36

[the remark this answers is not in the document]
nemesisdesign: openwisp-radius 1.0.0 is out and the other modules will be released very soon, so we'll pin them all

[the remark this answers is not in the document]
nemesisdesign: > I think this step will be right direction.
> 
> We will need to follow the release of dependency packages more closely. We can leverage "Depandabot" for that.

Yes, we should be using dependabot more, it's working pretty well in wifi-login-pages after we introduced browser testing.
 
> On a personal note, I am fed up of random failing builds.

Me too!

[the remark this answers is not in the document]
atb00ker: The only reason I was not pinning versions is because of the fact docker-openwisp is not in beta yet.
But if we are focusing on it, now, I think it'll be good to pin them! 😄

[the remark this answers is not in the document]
codesankalp: @atb00ker Is there any stable tag for `openwisp-base`?
I need to specify the tag in lines like `FROM openwisp/openwisp-base:latest` because using latest tag is giving warning.
I see few tags here: https://hub.docker.com/r/openwisp/openwisp-base/tags. Can you suggest which one to use except latest?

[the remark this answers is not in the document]
atb00ker: @codesankalp 
We can ignore that warning, because latest is build just before the images are build on local, so we want it to be `latest` only! 😄

nemesisdesign (changes requested):
@codesankalp all the modules have been released, you can proceed using all stable modules now.

images/openwisp_base/Dockerfile
```diff
@@ -7,67 +7,68 @@ FROM python:3.10-slim-bullseye AS SYSTEM
...
     fi
 ARG OPENWISP_UTILS_SOURCE=default
 RUN if [ "$OPENWISP_UTILS_SOURCE" != "default" ] ; then \
-    pip install --user --upgrade --force-reinstall ${OPENWISP_UTILS_SOURCE}; \
+    pip install --user --upgrade --force-reinstall ${OPENWISP_UTILS_SOURCE}~=1.0.1; \
     fi
 ARG DJANGO_X509_SOURCE=default
 RUN if [ "$DJANGO_X509_SOURCE" != "default" ]; then \
-    pip install --user --upgrade --force-reinstall ${DJANGO_X509_SOURCE}; \
+    pip install --user --upgrade --force-reinstall ${DJANGO_X509_SOURCE}~=1.1.0; \
```
nemesisdesign, line 64: These changes here are incorrect, it will break the possibility to install custom openwisp-modules.

Can you disable the linter only in this part of the dockerfile?

images/openwisp_base/Dockerfile
```diff
@@ -7,67 +7,68 @@ FROM python:3.10-slim-bullseye AS SYSTEM
...
 USER openwisp:root
 
 FROM SYSTEM AS PYTHON
 
 ENV PATH="${PATH}:/home/openwisp/.local/bin"
 ENV PYTHONPATH=/home/openwisp/.local/lib/python3.10/site-packages
+ENV EMAIL_BACKEND=djcelery_email.backends.CeleryEmailBackend
 
-RUN pip install --user --upgrade pip
-ARG OPENWISP_MONITORING_SOURCE=https://github.com/openwisp/openwisp-monitoring/tarball/docker-openwisp-usage
```
nemesisdesign, line 29: This should be:

``` 
ARG OPENWISP_MONITORING_SOURCE=openwisp-monitoring~=1.0.0
```

images/openwisp_base/Dockerfile
```diff
@@ -7,67 +7,68 @@ FROM python:3.10-slim-bullseye AS SYSTEM
...
 
-ARG DJANGO_SOURCE=django~=4.0.0
-RUN pip install --user --upgrade ${DJANGO_SOURCE}
+ARG DJANGO_SOURCE=django
+RUN pip install --user --upgrade ${DJANGO_SOURCE}~=4.0.0
 
 RUN if [ "$EMAIL_BACKEND" = "djcelery_email.backends.CeleryEmailBackend" ]; then \
-    pip install --prefix='/install' django-celery-email; \
+    pip install --user django-celery-email~=3.0.0; \
     fi
```
nemesisdesign, line 72: This one is good.

nemesisdesign (changes requested):
I see some changes in the base image Docker file which aren't what I expected, I see below.

images/openwisp_base/Dockerfile
codesankalp: Yes :+1:

codesankalp:
The tests are failing in fetching `https://pastebin.com/raw/ZMHRRYss` in openwisp-network-topology due to cloudflare check.

nemesisdesign (commented):
> The tests are failing in fetching https://pastebin.com/raw/ZMHRRYss in openwisp-network-topology due to cloudflare check.

Please add that file to this repo, under a new dir called `tests/static/'.

images/openwisp_nfs/init_command.sh
```diff
@@ -7,14 +7,13 @@
 set -ex
 
 mkdir -p $EXPORT_DIR/certs $EXPORT_DIR/postgres $EXPORT_DIR/static $EXPORT_DIR/media $EXPORT_DIR/html
-echo "$EXPORT_DIR   $EXPORT_OPTS" > /etc/exports
-
+echo "$EXPORT_DIR   $EXPORT_OPTS" >/etc/exports
 
 mount -t nfsd nfsd /proc/fs/nfsd
 # Fixed nlockmgr port
-echo 'fs.nfs.nlm_tcpport=32768' >> /etc/sysctl.conf
```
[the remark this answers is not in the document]
nemesisdesign: I am not sure I understand this, is it good to remove this line? And what did you mean for shell linting? I think this PR adds shell script linting too but I may have misunderstood your question.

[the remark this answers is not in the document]
atb00ker: Ah, please ignore! Deleting to avoid confusion!

images/openwisp_base/Dockerfile
```diff
@@ -7,67 +7,78 @@ FROM python:3.10-slim-bullseye AS SYSTEM
...
 RUN useradd --system --password '' --create-home --shell /bin/bash \
-   --gid root --uid 1001 openwisp
+    --gid root --uid 1001 openwisp
 USER openwisp:root
 
 FROM SYSTEM AS PYTHON
 
 ENV PATH="${PATH}:/home/openwisp/.local/bin"
 ENV PYTHONPATH=/home/openwisp/.local/lib/python3.10/site-packages
+ENV EMAIL_BACKEND=djcelery_email.backends.CeleryEmailBackend
```
atb00ker, line 27: Wait, why is this line here and repeated again with other env?
Does it belong here?

Wouldn't it make :
```

RUN if [ "$EMAIL_BACKEND" = "djcelery_email.backends.CeleryEmailBackend" ]; then \
    pip install --user django-celery-email~=3.0.0; \
    fi
```

always true? 

Even if we can pass the value ad-hoc to change `EMAIL_BACKEND`, I would say this should be by default not defined and we can check below:

```
if (EMAIL_BACKEND is undefined or EMAIL_BACKEND value != XYZ) 

THEN 
  do something
```

images/openwisp_base/Dockerfile
```diff
@@ -7,67 +7,78 @@ FROM python:3.10-slim-bullseye AS SYSTEM
...
 RUN if [ "$DJANGO_X509_SOURCE" != "default" ]; then \
     pip install --user --upgrade --force-reinstall ${DJANGO_X509_SOURCE}; \
     fi
 
-ARG DJANGO_SOURCE=django~=4.0.0
-RUN pip install --user --upgrade ${DJANGO_SOURCE}
+ARG DJANGO_SOURCE=django
+RUN pip install --user --upgrade ${DJANGO_SOURCE}~=4.0.0
 
 RUN if [ "$EMAIL_BACKEND" = "djcelery_email.backends.CeleryEmailBackend" ]; then \
```
atb00ker, line 80: Secondly, I wonder, why is this behind an `if` at all, can't we always install it like other deps?

images/openwisp_base/Dockerfile
```diff
@@ -7,67 +7,78 @@ FROM python:3.10-slim-bullseye AS SYSTEM
...
 ENV PATH="${PATH}:/home/openwisp/.local/bin"
 ENV PYTHONPATH=/home/openwisp/.local/lib/python3.10/site-packages
+ENV EMAIL_BACKEND=djcelery_email.backends.CeleryEmailBackend
 
-RUN pip install --user --upgrade pip
-ARG OPENWISP_MONITORING_SOURCE=https://github.com/openwisp/openwisp-monitoring/tarball/docker-openwisp-usage
+RUN pip install --user --upgrade pip~=22.0.4
+ARG OPENWISP_MONITORING_SOURCE=openwisp-monitoring~=1.0.0
+# hadolint ignore=DL3013
 RUN pip install --user --upgrade ${OPENWISP_MONITORING_SOURCE}
```
atb00ker, line 32: Now that we are using `--user`, I wonder,

We will have to copy `/home/openwisp/.local/` in the next step, would have increase the size of the final image 
OR would we be carrying any useless install dependency or system package like `gcc`'s residue that we don't need or want in the final image?


P.S: We specifically don't want `gcc` in the final image because it can potentially allow a malicious person to try to compile things, potentially making us vulnerable to a host of unknown exploits. (Just for your info, @codesankalp)

images/openwisp_base/Dockerfile
```diff
@@ -7,67 +7,78 @@ FROM python:3.10-slim-bullseye AS SYSTEM
...
     pip install --user --upgrade --force-reinstall ${OPENWISP_USERS_SOURCE}; \
     fi
 ARG OPENWISP_UTILS_SOURCE=default
+# hadolint ignore=DL3013
 RUN if [ "$OPENWISP_UTILS_SOURCE" != "default" ] ; then \
     pip install --user --upgrade --force-reinstall ${OPENWISP_UTILS_SOURCE}; \
     fi
 ARG DJANGO_X509_SOURCE=default
+# hadolint ignore=DL3013
 RUN if [ "$DJANGO_X509_SOURCE" != "default" ]; then \
```
atb00ker, line 73: Minor nitpick, can we add some space after if, like instead of:
```
if [ "$DJANGO_X509_SOURCE" != "default" ]; then \
pip install --user --upgrade --force-reinstall ${DJANGO_X509_SOURCE}~=1.1.0; \
fi
```

Can we do:

```
if [ "$DJANGO_X509_SOURCE" != "default" ]; then \
    pip install --user --upgrade --force-reinstall ${DJANGO_X509_SOURCE}~=1.1.0; \
fi
```

images/openwisp_base/Dockerfile
```diff
@@ -7,67 +7,78 @@ FROM python:3.10-slim-bullseye AS SYSTEM
...
     pip install --user --upgrade --force-reinstall ${DJANGO_X509_SOURCE}; \
     fi
 
-ARG DJANGO_SOURCE=django~=4.0.0
-RUN pip install --user --upgrade ${DJANGO_SOURCE}
+ARG DJANGO_SOURCE=django
+RUN pip install --user --upgrade ${DJANGO_SOURCE}~=4.0.0
```
atb00ker, line 78: What if I want to change the django version for my instance, how can I do that? :-O

atb00ker (changes requested):
Needs rebase btw.

images/openwisp_openvpn/Dockerfile
```diff
@@ -1,6 +1,7 @@
+# hadolint ignore=DL3007
 FROM kylemanna/openvpn:latest
 
-RUN apk add --no-cache tzdata postgresql-client supervisor && \
+RUN apk add --no-cache tzdata~=2022a-r0 postgresql-client~=12.10-r0  supervisor~=4.2.0-r0 && \
```
atb00ker, line 4: @nemesisdesign I forgot, what we concluded on locking these versions, in my experience, they don't change often, but when they do, we want to update them as quick as possible, mostly for security reasons.

atb00ker: Oh, wait, I remember now!
Using something like `requirements.io` and ensuring it can capture all our deps (preferably automatically) will ensure I can update them timely and specially if there is security update. 

@codesankalp , what do you think about that? Do you think that will solve the problem? Should we do it in this PR or a fast follow?

images/openwisp_base/Dockerfile
nemesisdesign: This part is broken @codesankalp

images/openwisp_base/Dockerfile
nemesisdesign: @atb00ker is this related to this PR or is it another subject altogether?If it's another subject, please open a new issue.

images/openwisp_base/Dockerfile
nemesisdesign: Please fix it @codesankalp

images/openwisp_base/Dockerfile
nemesisdesign: ok, I think we can simplify and just add `django-celery-email~=3.0.0` to the requirements.txt file

images/openwisp_openvpn/Dockerfile
nemesisdesign: I would do in a follow up PR, this one is becoming very time consuming to review

images/openwisp_base/Dockerfile
nemesisdesign: @atb00ker what dop you mean with "repeated again with other env"? Do you mean that it's already defined on line 103?

Looks weird. As wrote in a previous comment, we could simply consider django-celery-email a base requirement and install it anyway without doing all these checks.

nemesisdesign (changes requested):

images/openwisp_base/Dockerfile
codesankalp: > we could simply consider django-celery-email a base requirement and install it anyway without doing all these checks.

Moving forward with this because it will help in reducing complexity.

images/openwisp_base/Dockerfile
codesankalp: > it can potentially allow a malicious person to try to compile things, potentially making us vulnerable to a host of unknown exploits.

I didn't know about this. Thanks for letting me know.

I think we can open a separate issue or PR for this. WHat do you say?

images/openwisp_base/Dockerfile
codesankalp: > Secondly, I wonder, why is this behind an if

I think to make this change more like ansible-openwisp2 we added this conditional install.

Moving this into `requirements.txt` :+1:

images/openwisp_openvpn/Dockerfile
codesankalp: Yes we can use [snyk](https://snyk.io/) or something like that to update dependencies frequently.
I think doing this in another PR will be good.

nemesisdesign (approved):
LGTM.

@atb00ker will wait for your review, we may open new issues for following up with further improvements for dependency monitoring and multistage builds.

images/openwisp_openvpn/Dockerfile
atb00ker: Yes, makes sense!
Let's create an issue for this.

images/openwisp_base/Dockerfile
atb00ker: Potentially not related to this PR. Opening another issue!

PR #214, opened 2022-04-22
[feature] Added support for specifying custom Django settings #192

### LINKED ISSUE #192: [feature] Allow supplying custom settings
Allow to supply custom general python settings in order to customize any aspect of the application, like the SMTP settings or static storage settings needed to support cloud storage solutions.

pandafy:
Closes #192


.gitignore
```diff
@@ -80,3 +80,4 @@ local/
 
 # openwisp
 customize/
+!customize/configuration/custom_django_settings.example.py
```
nemesisdesign, line 83: double check this because `custom_django_settings.example.py` is missing

README.md
```diff
@@ -106,6 +107,14 @@ Setup on kubernetes is complex and requires prior knowledge about linux systems,
 
 ## Customization
 
+### Custom Django Settings
+
+If you want to add custom Django settings, you can follow the following guide.
+
+1. Create a directory `customize` in the same location as the `docker-compose.yml` file.
+2. Create a sub-directory `customize/configuration`.
+3. Create a file which will contain settings for Django `customize/configuration/custom_django_settings.py`. You can write all Django specific configuration in this file.
```
nemesisdesign, line 116: we can allow to add a directory instead of only one file and hence allow to add multiple python files to be loaded (which can be used to create custom apps or implement custom signal receivers).

docker-compose.yml
```diff
@@ -14,6 +14,7 @@ services:
       - openwisp_static:/opt/openwisp/static
       - openwisp_media:/opt/openwisp/media
       - influxdb_data:/var/lib/influxdb
+      - ./customize/configuration/custom_django_settings.py:/opt/openwisp/openwisp/custom_settings.py:ro
```
nemesisdesign, line 17:
```suggestion
      - ./customize/configuration:/opt/openwisp/openwisp/configuration:ro
```

Alterantively we could have something like:

/configuration/theme
/configuration/django

or 

/customization/theme
/customization/django

build/common/openwisp/settings.py
```diff
@@ -386,3 +386,8 @@
         INSTALLED_APPS.remove('openwisp_monitoring.check')
 if EMAIL_BACKEND == 'djcelery_email.backends.CeleryEmailBackend':
     INSTALLED_APPS.append('djcelery_email')
+
+try:
+    from .custom_settings import *
```
nemesisdesign, line 391: This will have to be updated according to the new directory structure.
```python
from .customization.django.settings.py import *
```

customize/configuration/custom_django_settings.py
```diff
@@ -0,0 +1,2 @@
+# This file is required to run the test project.
+# DO NOT COMMIT CHANGES TO THIS FILE.
```
nemesisdesign, line 2: We should also be able to remove this and add it to the git ignore.

We should provide an empty `__init__.py` file as well.

README.md
nemesisdesign: I think we need an instruction like:

```
cp <path>/custom_django_settings.example.py <path>/custom_django_settings.py
```

Eg:

```
cp customization/django/settings.example.py customization/django/settings.py
```

nemesisdesign (changes requested):
Good progress :+1: 
See below.

nemesisdesign (approved):
Thanks! :+1:

PR #236, opened 2022-06-14
[feature] Added django-storage and boto3 in openwisp-base image #233

### LINKED ISSUE #233: [feature] Add django-storage and boto3 to requirements.txt of base image
These libraries are really widely used to implement object storage support and we should include them by default in the base images used for the django apps.

pandafy:
Closes #233

### DIFF
--- a/images/openwisp_base/requirements.txt
+++ b/images/openwisp_base/requirements.txt
@@ -8,3 +8,5 @@ django-cors-headers~=3.1  # allows 3.x and > 3.1
 django-pipeline~=2.0  # allows 2.x
 uwsgi~=2.0.20
 django-celery-email~=3.0.0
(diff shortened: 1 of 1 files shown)


images/openwisp_base/requirements.txt
```diff
@@ -8,3 +8,5 @@ django-cors-headers~=3.1  # allows 3.x and > 3.1
 django-pipeline~=2.0  # allows 2.x
 uwsgi~=2.0.20
 django-celery-email~=3.0.0
+django-storages~=1.12.3
```
nemesisdesign, line 11: let's at least add a comment which explains why these are here

nemesisdesign (changes requested):

Issue #254, opened 2022-09-08
[Bug] The base image has an invalid stage name


scott-grimes:
building the base image yields:

```
[+] Building 0.3s (2/2) FINISHED                                                                                              
 => [internal] load build definition from Dockerfile                                                                     0.2s
 => => transferring dockerfile: 5.62kB                                                                                   0.0s
 => [internal] load .dockerignore                                                                                        0.1s
 => => transferring context: 2B                                                                                          0.0s
failed to solve with frontend dockerfile.v0: failed to create LLB definition: failed to parse stage name "SYSTEM": invalid reference format: repository name must be lowercase
```

PR #255, opened 2022-09-08
lowercase stage name for openwisp base image

scott-grimes:
Resolves issue https://github.com/openwisp/docker-openwisp/issues/254


nemesisdesign (approved):
Thanks! :+1:

PR #260, opened 2022-12-09
[fix] Replaced postfix `hash:` with `lmdb:`

Bixilon:
This fixes a crash caused by upgrading the postifix alpine base image in 319ff0568b437c0d60d603294853683e76fb0d40. Looks like somebody did not test their changes :)

(I just tested it with a dummy postfix backend, but should still do the trick)

### DIFF
--- a/images/common/utils.sh
+++ b/images/common/utils.sh
@@ -148,9 +148,9 @@ function postfix_config {
 
 	postconf -e myhostname="$POSTFIX_MYHOSTNAME"
 	postconf -e myorigin='$myhostname'
-	postconf -e alias_maps=hash:/etc/aliases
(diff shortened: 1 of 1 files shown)


Aryamanz29:
Produced Logs, Ref here : https://gitlab.alpinelinux.org/alpine/aports/-/issues/12381
```bash
root@arya-vm:/home/arya# docker logs -f 597122812fb2 
postalias: fatal: unsupported dictionary type: hash. Is the postfix-hash package installed?
postalias: fatal: unsupported dictionary type: hash. Is the postfix-hash package installed?

```

Aryamanz29:
Thanks for contributing @Bixilon!

Closed by: https://github.com/openwisp/docker-openwisp/pull/271

PR #271, opened 2023-03-15
[fix] Fixed docker-openwisp installation

Aryamanz29:
This PR fixes:

#### Containers

- `openwisp-freeradius`
- `openwisp-postifix`
- `openwisp-nginx`
- `openwisp-dashboard` (mainly autocomplete filter errors)

#### Others

- Docker tests and QA checks

Screenshot from 2023-03-15 18-30-50

Screenshot from 2023-03-15 17-01-04

### DIFF
--- a/images/openwisp_freeradius/Dockerfile
+++ b/images/openwisp_freeradius/Dockerfile
@@ -1,7 +1,7 @@
 FROM freeradius/freeradius-server:3.0.26-alpine
 
 # hadolint ignore=DL3018
-RUN apk add --no-cache --update tzdata postgresql-dev~=13.8-r0 \
+RUN apk add --no-cache --update tzdata~=2022f-r1 postgresql-dev~=13.8-r0 \
                                 postgresql-client~=13.8-r0 && \
     rm -rf /var/cache/apk/* /tmp/*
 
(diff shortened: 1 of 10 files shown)


images/openwisp_freeradius/Dockerfile
```diff
@@ -1,6 +1,7 @@
-FROM freeradius/freeradius-server:3.0.25-alpine
+FROM freeradius/freeradius-server:3.0.26-alpine
 
-RUN apk add --no-cache --update tzdata~=2021c-r0 postgresql-dev~=13.8-r0 \
+# hadolint ignore=DL3018
+RUN apk add --no-cache --update tzdata postgresql-dev~=13.8-r0 \
```
nemesisdesign, line 4: why was the pinning removed here?

Bixilon: because apline is publishing new tzdata packages and the old ones are not available anymore (I suggested that)

Bixilon (approved):

images/openwisp_freeradius/Dockerfile
Bixilon: That was not what I meant. I mena, that (with the same version of alpine). I am not using alpine often, but as far as I have seen (here), the tzdata package version got incremented (e.g. from `d` to `e`) withing the same version alpine. That broke the build after a while. 

I think we should pin dependencies (it is always good for reproducible builds and nothing should break), but some packages (like `tzdata`) will never break the system within the same version of alpine (and probably not even when updating alpine). We don't explicitly depend on that package (I mean we don't use their apis in a way, where it could break), it is more like a general thing to have. imho we don't need to pin such dependencies, we only might miss fixes security holes, because of manual updating.

nemesisdesign (approved):

Issue #272, opened 2023-03-16
[bug] Publishing to container registry is broken


Bixilon:
So, the latest commit (29d564a33f7364f9c5cff84be1fa9030f0ad24d0) fails:

Publishing to docker hub does not work, because of: `denied: requested access to the resource is denied`

Also publishing to gitlab container registry does not work (unauthorized).

I suggest the following things:
 - [ ] Make pipeline fail when publishing to docker hub fails
 - [ ] Fix docker hub publishing @nemesisdesign 
 - [ ] Don't publish to gitlab container registry anymore (tha should then fix ci)
 - [ ] remove all references to gitlab container registry (docker compose, makefile, ...)

nemesisdesign:
Closed by https://github.com/openwisp/docker-openwisp/commit/ef5284e3f58938981c3deaefc923f357221da569.
We're not going to remove the gitlab registry because dockerhub has discontinued the free teams for open source project in favour of another open source offering which we need to apply for, but we already have a similar offering with gitlab which provides more features so we will most likely remove dockerhub at this point.

Bixilon:
Then do that, instead of having docker hub images, people might be confused. Nobody looks at the gitlab registry if something is on docker hub.

PR #277, opened 2023-03-20
[change] Remove docker hub publishing

Bixilon:
Removes docker hub completely to avoid confusion.

I also felt like renaming the secret `DOCKER_USER` to `GITLAB_DOCKER_REGISTRY_USER` is appropriate. Somebody of you needs to to that, before merging.

See #272


nemesisdesign:
Thank you for your collaboration @Bixilon, but taking care of docs and CI is the maintainer's job and must be discussed before being worked out. I recommend focusing your effort on specific technical issues, that's where your help will be most effective.

PR #278, opened 2023-03-20
Cleanup readme

Bixilon:
I moved (without reading through and checking the content) a lot of information from the readme to the `docs` folder.

Users normally don't want to be bloated about information that is not relevant for them. They just want to know how it set it up and use it (maybe we should add a dashboard screenshot too!).


nemesisdesign:
Thank you for your collaboration @Bixilon, but taking care of docs and CI is the maintainer's job and must be discussed before being worked out. I recommend focusing your effort on specific technical issues, that's where your help will be most effective.

PR #281, opened 2023-03-21
[fix] Fix domain detection

### LINKED ISSUE #273: [bug] device charts are not loading
Device charts refuse to load, because of:
`jquery.min.8fb8fee4fcc3.js:2 Refused to connect to 'https://openwisp-api.xx.muc/api/v1/monitoring/device/886d42ca-7fc6-4282-b6e5-a4a5b5e17c7f/?key=key&timezone=Europe/Berlin&time=7d' because it violates the following Content Security Policy directive: "connect-src *. wss: 'self'"`

Adding a security policy header to nginx should fix that issue

We need to add the dashboard domain in `/images/common/utils.sh` + add it to the config if needed, see `/etc/nginx/openwisp.ssl.template.conf`

Bixilon:
Should fix #273, not tested

### DIFF
--- a/images/openwisp_nginx/get_domain.py
+++ b/images/openwisp_nginx/get_domain.py
@@ -1,7 +1,4 @@
 import os
 import sys
 
-import tldextract
-
-ext = tldextract.extract(os.environ['API_DOMAIN'])
-sys.stdout.write(ext.registered_domain)
+sys.stdout.write(".".join(os.environ['API_DOMAIN'].split(".")[1:]))
--- a/images/openwisp_nginx/requirements.txt
+++ b/images/openwisp_nginx/requirements.txt
@@ -1,1 +1,0 @@
-tldextract~=3.2.0


nemesisdesign (commented):
Can you explain this change with a few examples of why the current way is not working and the new way is better, please?

Bixilon:
I have no clue what the library is doing, but it should cut off the first part of a domain (`test.abc.de´ -> `abc.de`). The library just does it with the top level domain (`test.abc.de` -> `de`, but not even then correct, check a domain with a dash `test.ab-cd.de`, it then just returns nothing).

It is used in a header (nginx), and if that header is not set correctly, things like chart loading fail.

Bixilon:
Tested it, working fine!

Screenshot_20230324_162010

Bixilon:
Any chance of getting this merged?

nemesifier:
Testing with `tldextract-5.1.2`:

```python
In [9]: print(tldextract.extract('openwisp.org').registered_domain)
openwisp.org

In [10]: print(tldextract.extract('api.openwisp.org').registered_domain)
openwisp.org

In [11]: print(tldextract.extract('test.api.openwisp.org').registered_domain)
openwisp.org

In [12]: print(tldextract.extract('test.abc.de').registered_domain)
abc.de

In [13]: print(tldextract.extract('abc.de').registered_domain)
abc.de

In [14]: print(tldextract.extract('test.test.abc.de').registered_domain)
abc.de
```

I cannot replicate the issue you are mentioning here. I will try to update this library and modify the different parts of the code to make sure they're consistent.

Bixilon:
try extracting from api.open-wisp.de

nemesifier:
> try extracting from api.open-wisp.de

As far as I know, dashes are not allowed in the domain

> try extracting from api.open-wisp.de

```
In [1]: import tldextract

In [2]: ext=tldextract.extract('api.open-wisp.org')

In [3]: ext
Out[3]: ExtractResult(subdomain='api', domain='open-wisp', suffix='org', is_private=False)

In [4]: ext.registered_domain
Out[4]: 'open-wisp.org'
```

What's the issue? Please update your local copy of the library.
PR in https://github.com/openwisp/docker-openwisp/pull/300.

PR #283, opened 2023-05-21
Some updates

katomaso:
- update versions of openwisp-* to the latest
- get rid of openwisp user in docker images
- put a timeout in dashboard service discovery
- rearrange images to be a bit more effective

### DIFF
--- a/images/common/openwisp/settings.py
+++ b/images/common/openwisp/settings.py
@@ -58,6 +58,8 @@ CORS_ALLOWED_ORIGINS = [
     f'{HTTP_SCHEME}://{os.environ["API_DOMAIN"]}',
 ] + os.environ['DJANGO_CORS_HOSTS'].split(',')
 CORS_ALLOW_CREDENTIALS = True
+CORS_ALLOW_ALL_ORIGINS = DEBUG
+CSRF_TRUSTED_ORIGINS = CORS_ALLOWED_ORIGINS
 
 if HTTP_SCHEME == 'https':
     SESSION_COOKIE_SECURE = True
@@ -298,7 +300,7 @@ LOGGING = {
             'stream': sys.stdout,
         },
         'mail_admins': {
-            'level': os.environ['DJANGO_LOG_LEVEL'],
+            'level': 'ERROR',
             'class': 'django.utils.log.AdminEmailHandler',
             'filters': ['require_debug_false', 'user_filter'],
         },
--- a/images/openwisp_base/Dockerfile
+++ b/images/openwisp_base/Dockerfile
@@ -8,13 +8,10 @@ FROM python:3.10-slim-bullseye AS system
 RUN apt-get update && \
     apt-get install --yes --no-install-recommends \
     libcairo2 apt-utils libpangocairo-1.0-0 \
-    gdal-bin gettext fping && \
-    rm -rf /var/lib/apt/lists/* /root/.cache/pip/* /tmp/*
-# hadolint ignore=DL3008, DL3009
-RUN apt-get update && \
-    apt-get install --yes --no-install-recommends \
+    gdal-bin gettext fping \
     gcc libpq-dev libjpeg-dev libffi-dev python3-dev \
-    python3-pip libxml2-dev libxslt1-dev zlib1g-dev g++ procps
+    python3-pip libxml2-dev libxslt1-dev zlib1g-dev g++ procps git && \
+    rm -rf /var/lib/apt/lists/* /root/.cache/pip/* /tmp/*
 
 RUN useradd --system --password '' --create-home --shell /bin/bash \
     --gid root --uid 1001 openwisp
@@ -25,7 +22,7 @@ FROM system AS openwisp_python
 ENV PATH="${PATH}:/home/openwisp/.local/bin"
 ENV PYTHONPATH=/home/openwisp/.local/lib/python3.10/site-packages
 
-RUN pip install --no-cache-dir  --user --upgrade pip~=23.0.1
+RUN pip install --no-cache-dir  --user --upgrade pip wheel
 # TODO: Remove when next version of openwisp-monitoring is released
 ARG OPENWISP_MONITORING_SOURCE=https://github.com/openwisp/openwisp-monitoring/tarball/15977c74b48d846ac469316abc7f13ccbef405f0
 # hadolint ignore=DL3013
--- a/images/openwisp_base/Dockerfile
+++ b/images/openwisp_base/Dockerfile
@@ -8,90 +8,86 @@ FROM python:3.10-slim-bullseye AS system
 RUN apt-get update && \
     apt-get install --yes --no-install-recommends \
     libcairo2 apt-utils libpangocairo-1.0-0 \
-    gdal-bin gettext fping \
-    gcc libpq-dev libjpeg-dev libffi-dev python3-dev \
-    python3-pip libxml2-dev libxslt1-dev zlib1g-dev g++ procps git && \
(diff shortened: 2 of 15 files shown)


images/openwisp_freeradius/Dockerfile
```diff
@@ -1,8 +1,7 @@
-FROM freeradius/freeradius-server:3.0.26-alpine
+FROM freeradius/freeradius-server:3.2.2-alpine
 
 # hadolint ignore=DL3018
-RUN apk add --no-cache --update tzdata~=2022f-r1 postgresql-dev~=13.8-r0 \
-                                postgresql-client~=13.8-r0 && \
+RUN apk add --no-cache --update tzdata postgresql-dev postgresql-client && \
```
Bixilon, line 4: why unstick the depencies?

images/openwisp_base/Dockerfile
```diff
@@ -8,93 +8,86 @@ FROM python:3.10-slim-bullseye AS system
 RUN apt-get update && \
     apt-get install --yes --no-install-recommends \
     libcairo2 apt-utils libpangocairo-1.0-0 \
-    gdal-bin gettext fping && \
+    gdal-bin gettext fping nano && \
```
atb00ker, line 11: Why do we need an editor in production images?
Did you add this during testing?

images/openwisp_websocket/Dockerfile
```diff
@@ -1,12 +1,13 @@
 # hadolint ignore=DL3007
-FROM openwisp/openwisp-base:latest
+ARG BASE_IMAGE=openwisp/openwisp-base:latest
+FROM ${BASE_IMAGE}
 
 WORKDIR /opt/openwisp/
 
-COPY --chown=openwisp:root  ./common/ \
```
atb00ker, line 6: Why are we removing this?
If I remember correctly, the intention of making `openwisp` the owner was because we did not want to run things as root in production.

images/openwisp_base/Dockerfile
```diff
@@ -8,93 +8,86 @@ FROM python:3.10-slim-bullseye AS system
...
 
-COPY --from=openwisp_python --chown=openwisp:root /home/openwisp/.local/ /usr/local
-COPY --chown=openwisp:root ./common/ /opt/openwisp/
-RUN mkdir /opt/openwisp/static && mkdir /opt/openwisp/media && mkdir /opt/openwisp/private && \
-    mkdir /opt/openwisp/logs && chown -R openwisp:root /opt/openwisp
+COPY --from=openwisp_python /usr/local/lib/python3.10/site-packages /usr/local/lib/python3.10/site-packages
+COPY --from=openwisp_python /usr/local/bin /usr/local/bin
+COPY ./common/ /opt/openwisp/
+RUN mkdir /opt/openwisp/static && mkdir /opt/openwisp/media && mkdir /opt/openwisp/private && mkdir /opt/openwisp/logs && \
+    chmod a+w /opt/openwisp /opt/openwisp/static /opt/openwisp/media /opt/openwisp/private /opt/openwisp/logs
```
atb00ker, line 81: Answers my permission question I guess; I wonder if `a+r` would be enough or not.

images/common/openwisp/settings.py
```diff
@@ -298,7 +300,7 @@
             'stream': sys.stdout,
         },
         'mail_admins': {
-            'level': os.environ['DJANGO_LOG_LEVEL'],
+            'level': 'ERROR',
```
atb00ker, line 303: Why is the variable removed?

images/openwisp_websocket/Dockerfile
katomaso: Hello, thank you for the feedback. Internal user is great until you try to bind volumes from host's system (our case). OpenWISP user has fixed UID 1001 that might mean something on the host and it will mess things up. It is much preferable to specify container's runtime user either via `UID` and `GID` variables or via `user` option in docker(compose). The first approach is used in all `bitnami` images but I am a big fan of the later. But both works in this case.

images/openwisp_base/Dockerfile
katomaso: Unfortunately no because of `collectstatics` that is run at the start and might write to `/opt/openwisp/static` and optionally, the webserver might write to ``/opt/openwisp/media` folder. uWSGIs write to `/opt/openwisp/log` which could be changed to `/var/log/openwisp` wit correct persmissions but I didn't want to do too many changes. And finally `/opt/openwisp/` itself is used to render `uwsgi.conf` during runtime. This could be modified by using `init_command.sh` as the `ENTRYPOINT` and pass parameters to uwsgi via `CMD` array. But again, this would result in even more changes.

images/common/openwisp/settings.py
katomaso: This I was just unsure. If I set DJANGO_LOG_LEVEL to "INFO" ... would it send me email with every INFO log record?

images/openwisp_freeradius/Dockerfile
katomaso: Hi Bixilion, I don't have a good answer for that. We are running psql15 so I just wanted as new client as possible.

images/openwisp_base/Dockerfile
katomaso: Well my reasoning was that since the containers are running unprivileged then getting an editor there is not very straightforward in case something breaks in production. 80kB or how much it weights sounds like a meaningful sacrifice. Given that the image itself has over 1.1GB

images/openwisp_freeradius/Dockerfile
nemesisdesign: we should not do this, we have run into a lot of trouble with unsticked dependencies: the building process becomes very unstable, it breaks periodically because some new dependency introduces changes which do not play well with the process.
The upgrade of the dependency has to be done periodically.

images/openwisp_websocket/daphne.conf
```diff
@@ -1,12 +1,10 @@
 [supervisord]
 nodaemon=true
-user=openwisp
```
nemesisdesign, line 3: why? This was done on purpose to not run the service as root.

images/common/openwisp/settings.py
nemesisdesign: You can change this from the env, no need to hard code it:
https://github.com/openwisp/docker-openwisp/blob/master/.env#L47

nemesisdesign: It won't send you any email unless you define the admins email in the settings, consult the django documentation for more info: https://docs.djangoproject.com/en/4.2/ref/settings/#admins

nemesisdesign (changes requested):
Thank you for your contribution @katomaso, we'll review the changes.

I recommend going easy with the changes, a more gradual approach is recommended on my side.

Please see my comments below.

images/openwisp_freeradius/Dockerfile
katomaso: I have no strong opinion on that so I reverted my changes.

images/common/openwisp/settings.py
nemesisdesign: In this case I think the best is to add another env variable for this specific line, e.g.:

```python
os.environ['DJANGO_MAIL_ADMINS_LOG_LEVEL']
```

I agree it doens't make sense to send info via email. It looks to me that this feature of sending events via email is just not used at all. I personally use sentry or other log collectors so I don't even bother with this but it would be great to fix it.

The reason for not hardcoding is that here 2 possible values make sense: ERROR and CRITICAL.

Issue #241, opened 2022-06-18
[bug] Django warnings in API container


nemesisdesign:
API container gives this warning when using `python manage.py`:

?: (urls.W005) URL namespace 'firmware' isn't unique. You may not be able to reverse all URLs in this namespace
?: (urls.W005) URL namespace 'ipam' isn't unique. You may not be able to reverse all URLs in this namespace

Vihar214:
Hello, Is there anyone working on this issue if not then can you assign this issue to Me?

uzaxirr:
Hey @nemesisdesign 

I would like to work on this, The warning means that multiple URL routes are having the same namespace. I'll have to Assign a unique namespace to each URL. But can you tell me which container is generating this warning so that i can figure out the appropriate repo to make change in.

nemesisdesign:
@uzaxirr try spawning the containers with `make develop` and you should be able to find this on your own.
@Vihar214 we don't assign issues to non core contributors, just announce you're working on something, open your PRs and ask for feedback.

If you're keen on contributing, please join [our dev chat](https://gitter.im/openwisp/development) to coordinate.

uzaxirr:
> @uzaxirr try spawning the containers with `make develop` and you should be able to find this on your own. @Vihar214 we don't assign issues to non core contributors, just announce you're working on something, open your PRs and ask for feedback.
> 
> If you're keen on contributing, please join [our dev chat](https://gitter.im/openwisp/development) to coordinate.

I'm unable to re-produce this, I did make develop and spinned up the containers, [here](https://app.warp.dev/block/31mkm7clEZF3fQOZjkzC8y) is the output

tinman280:
Can you assign me this issue?

Adityashandilya555:
working on this issue

PR #288, opened 2023-09-27
[fix] django namespace error in api container #241

shubham-yadavv:
django namespace error in api container for the firmaware URL namespace

### DIFF
--- a/images/openwisp_api/urls.py
+++ b/images/openwisp_api/urls.py
@@ -27,7 +27,7 @@ if env_bool(os.environ['USE_OPENWISP_FIRMWARE']):
             'api/v1/',
             include('openwisp_firmware_upgrader.api.urls', namespace='firmware'),
         ),
-        path('', include((fw_private_storage_urls, 'firmware'), namespace='firmware')),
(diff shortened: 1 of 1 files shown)


images/openwisp_api/urls.py
```diff
@@ -27,7 +27,7 @@
             'api/v1/',
             include('openwisp_firmware_upgrader.api.urls', namespace='firmware'),
         ),
-        path('', include((fw_private_storage_urls, 'firmware'), namespace='firmware')),
+        path('', include((fw_private_storage_urls, 'firmware'), namespace='firmware-private')),
```
nemesifier, line 30: doesn't look consistent with what we have in ansible-openwisp2:

https://github.com/openwisp/ansible-openwisp2/blob/master/templates/openwisp2/urls.py#L26-L32

nemesifier (changes requested):

nemesifier:
I am closing this for now because I do not understand what problem this patch is addressing.
If you think I am wrong, please open a bug report indicating what actions you're performing, the expected result and the actual wrong result you're getting.
Thanks for contributing!

PR #338, opened 2024-07-26
[fix] Fixed URL for downloading CRL from OpenWISP

### LINKED ISSUE #231: [bug] OpenVPN tunnels does not work out of the box
**Steps to replicate**
1. Start the development environment 
2. Register a device to the development environment. The default VPN template should get automatically applied to the device. 

**Expected Outcome**
An OpenVPN tunnel is established between device and the OpenVPN container. And it is possible to send traffic through it

**Actual Outcome**
OpenVPN on the router fails with following error
```
Thu Jun  9 17:23:26 2022 daemon.err openvpn(default)[9866]: TLS Error: TLS key negotiation failed to occur within 60 seconds (check your network connectivity)

```

pandafy:
Set tls-timeout to 2 for OpenVPN server

Closes #231 

--------------

The OpenVPN image in docker-openwisp is configured to download the CRL from the OpenWISP dashboard. The URL mentioned in the script (in docker-openwisp) is wrong. Thus, the downloaded CRL files contains HTML (due to redirect to login page) and the OpenVPN server runs into error while reading that.


images/common/utils.sh
```diff
@@ -237,5 +237,5 @@ function openvpn_config_download {
 
 function crl_download {
 	export CAid=$(psql -qAtc "SELECT ca_id FROM config_vpn where name='${VPN_NAME}';")
-	wget -qO revoked.crl --no-check-certificate ${DASHBOARD_INTERNAL}/admin/pki/ca/${CAid}.crl
+	wget -qO revoked.crl --no-check-certificate ${DASHBOARD_INTERNAL}/admin/pki/ca/x509/ca/${CAid}.crl
```
pandafy, line 240: We will need to update this once https://github.com/openwisp/django-x509/pull/147 is merged.

nemesifier (approved):

PR #362, opened 2024-09-18
[deps] Fixed openssl dependency in Nginx Image #318

pandafy:
## Checklist

## Description of Changes

In #318, I initially removed the OpenSSL dependency from the Nginx image, assuming it was included in the base image. However, the base image removes the OpenSSL package in a later build step. This commit reinstates the OpenSSL dependency to ensure proper functionality.

Related to #318

### DIFF
--- a/images/openwisp_nginx/Dockerfile
+++ b/images/openwisp_nginx/Dockerfile
@@ -1,6 +1,7 @@
 FROM nginx:1.27.0-alpine
 
 RUN apk add --update --no-cache \
+            openssl~=3.1.7-r0 \
             py3-pip~=23.3.1-r0 \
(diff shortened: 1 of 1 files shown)


pandafy:
I investigated why this issue was not caught by the CI and discovered that the [`openssl` command is only triggered when the SSL certificate and key are missing](https://github.com/openwisp/docker-openwisp/blob/19aabb5111be2e2094126895a98e944ada3afcd1/images/common/utils.sh#L51-L61). In our CI process, we first deploy the existing edge version before deploying the image built in the CI. As a result, the certificates were already present in the second deployment, so the `openssl` command was never executed, allowing the bug to go unnoticed.

nemesifier (approved):

PR #460, opened 2025-05-08
[chores] Migrated to Django 4.2+ STORAGES setting #442

nemesifier:
Related to #442.


images/common/openwisp/__init__.py
```diff
@@ -3,5 +3,5 @@
 from .celery import app as celery_app
 
 __all__ = ['celery_app']
-__openwisp_version__ = '24.11.1'
+__openwisp_version__ = '25.06.0a'
```
pandafy, line 6: @nemesifier what do you think about the version?

PR #473, opened 2025-06-07
[feature] Run collectstatic only when python dependencies change #246

### LINKED ISSUE #246: [feature] Execute collect static only when versions change
I have an idea of how we could deal with extreme slowness of calling collectstatic when using object storage:

- We could hash the output of `pip freeze` in the dashboard container
- Before calling `collectstatic` we check in redis to see if the current pip freeze hash is different from the one  saved there
- If it's different we call collectstatic and then save the new hash in redis

@pandafy @atb00ker what do you think?

stktyagi:
## Checklist

Closes #246

## Description of Changes

Added a collectsatic_control.py file that computes a hash of the current installed Python packages (pip freeze output) and connects to Redis to retrieve a previously stored hash. If the hash is missing or changed, it runs Django’s collectstatic command to gather static files and updates the stored hash in Redis and if the hash is unchanged, it skips running collectstatic.


images/common/init_command.sh
```diff
@@ -14,7 +14,7 @@ if [ "$MODULE_NAME" = 'dashboard' ]; then
 	python manage.py migrate --noinput
 	test -f "$SSH_PRIVATE_KEY_PATH" || ssh-keygen -t ed25519 -f "$SSH_PRIVATE_KEY_PATH" -N ""
 	python load_init_data.py
-	python manage.py collectstatic --noinput
+	python /opt/openwisp/collectstatic_control.py
```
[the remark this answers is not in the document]
stktyagi: was running into some issues with this but turns out it was a small setup issue on my system. Nevertheless I have fixed and used relative path in latest commit.

images/common/init_command.sh
```diff
@@ -14,7 +14,7 @@ if [ "$MODULE_NAME" = 'dashboard' ]; then
 	python manage.py migrate --noinput
 	test -f "$SSH_PRIVATE_KEY_PATH" || ssh-keygen -t ed25519 -f "$SSH_PRIVATE_KEY_PATH" -N ""
 	python load_init_data.py
-	python manage.py collectstatic --noinput
+	python collectstatic_control.py
```
nemesifier, line 17: Let's simplify.

```suggestion
	python collectstatic.py
```

images/common/collectstatic_control.py
```diff
@@ -0,0 +1,38 @@
+import hashlib
```
nemesifier, line 1: Let's add a docstring at the top resuming what this file is used for, eg:

```
""""
Calls `collectstatic` only when dependencies have changed.
This helps speed up the startup sequence when using
public cloud providers.

This behavior can be disabled by setting the env var
`COLLECTSTATIC_WHEN_DEPS_CHANGE` to `False`.
""""
```

nemesifier (commented):
@stktyagi great progress here!

Can you please add an env var called `COLLECTSTATIC_WHEN_DEPS_CHANGE`, defaulting to `True`, which allows us to disable this feature?
This can be useful during development or on systems where we  are using the
development version and the version identifier of the openwisp modules does not change but we still need to call `collectstatic` anyway.

We'll need to add this var to the docs as well.

stktyagi:
There was some conflict between formatter and the qa tester while adding the docstring in collecstatic, hence the redundant commits. Will make sure in the future.

images/common/collectstatic.py
```diff
@@ -0,0 +1,47 @@
...
+        pip_freeze_output = subprocess.check_output(["pip", "freeze"]).decode()
+        return hashlib.sha256(pip_freeze_output.encode()).hexdigest()
+    except subprocess.CalledProcessError as e:
+        print(f"Error running 'pip freeze': {e}", file=sys.stderr)
+        sys.exit(1)
+
+
+def main():
+    if os.environ.get("COLLECTSTATIC_WHEN_DEPS_CHANGE", "true").lower() == "false":
+        print("COLLECTSTATIC_WHEN_DEPS_CHANGE is false; skipping collectstatic.")
```
nemesifier, line 31: This is not what I meant.
If the env var is False we shall always execute `collectstatic`. **Not executing it is not an option**.

nemesifier (changes requested):

nemesifier (approved):
I did some minor improvements, feel free to review. LGTM :heavy_check_mark:

stktyagi:
> I did some minor improvements, feel free to review. LGTM ✔️

looks like I forgot to update the docs, my bad. Thank you for doing that.

PR #555, opened 2026-01-24
[fix] Added DOCKER_TAG variable for version pinning to match intended release version

### LINKED ISSUE #554: [bug] docker-compose uses 'latest' tags; should use version numbers
**Describe the bug**
Using `latest` tags in docker-compose can cause unwanted upgrades if using `docker compose pull` instead of `make pull` (which pulls the correct image and tag them as latest). This may accidentally pull incompatible versions and break a working deployment unexpectedly. This is especially risky for users running older or stable releases (for example, 24.11), because pulling will download newer images (e.g., 25.10.1) without warning.

**Steps To Reproduce**
1. Deploy using docker-compose for a specific release (e.g., 24.11)
2. Run `docker compose pull`
3. Docker downloads images tagged as `latest` (typically the newest release)
4. Deployment may break due to incompatibility or unexpected updates

**Expected behavior**
Images should always match the intended release version in the compose file, avoiding accidental upgrades or mismatches regardless of command usage.

**Screenshots**
N/A

**System Informatioon:**
- OS: Any
- Docker version: Any
- Browser and Browser Version (if applicable): N/A
- Additional context: The workflow currently relies on `make pull` to determine and overwrite the tags as `latest`, but this is not a reliable safeguard against accidental upgrades.

atif09:
this change introduces a DOCKER_TAG environment variable that allows users to pin specific image versions in .env file, both 'docker compose pull' and 'make pull' now work according to this variable, ensuring consistent version behavior across all deployment methods

## Checklist

Closes #554.

## Description of Changes

- Add DOCKER_TAG=latest to .env file
- Update all OpenWISP image tags in docker-compose.yml to use ${DOCKER_TAG:-latest}
- Update Makefile to include .env and use DOCKER_TAG when retagging images

## Screenshot
version pinning test:

**with `DOCKER_TAG=25.10.0` in .env**
image

### DIFF
--- a/deploy/auto-install.sh
+++ b/deploy/auto-install.sh
@@ -59,7 +59,7 @@ apt_dependenices_setup() {
 }
 
 get_version_from_user() {
-	echo -ne ${GRN}"OpenWISP Version (leave blank for latest stable release): "${NON}
+	echo -ne "${GRN}OpenWISP Version (leave blank for latest stable release): ${NON}"
 	read openwisp_version
 	if [[ -z "$openwisp_version" ]]; then
 		openwisp_version=$(curl -L --silent https://api.github.com/repos/openwisp/docker-openwisp/releases/latest | jq -r .tag_name)
--- a/docs/user/settings.rst
+++ b/docs/user/settings.rst
@@ -37,7 +37,7 @@ Essential
 You will need to adapt these values to get the docker images working
 properly on your system.
 
-.._openwisp_version:
+.. _openwisp_version:
 
 ``OPENWISP_VERSION``
 ~~~~~~~~~~~~~~~~~~~~


Makefile
```diff
@@ -1,6 +1,9 @@
 # Find documentation in README.md under
 # the heading "Makefile Options".
 
+include .env
+export
+
```
atif09: ive added a blank line here which I'll delete

Makefile
```diff
@@ -20,7 +23,7 @@ pull:
 				 'openwisp-freeradius' 'openwisp-nginx' 'openwisp-openvpn' 'openwisp-postfix' \
 				 'openwisp-websocket' ; do \
 		docker pull --quiet $(USER)/$${image}:$(TAG); \
-		docker tag  $(USER)/$${image}:$(TAG) openwisp/$${image}:latest; \
+		docker tag  $(USER)/$${image}:$(TAG) openwisp/$${image}:$${DOCKER_TAG:-latest}; \
```
pandafy: This means, we will always have versioned tagged images.

deploy/auto-install.sh
```diff
@@ -87,7 +87,7 @@ download_docker_openwisp() {
 		rm -rf $INSTALL_PATH &>>$LOG_FILE
 	fi
 	if [ -z "$GIT_BRANCH" ]; then
-		if [[ "$openwisp_version" == "edge" ]]; then
+		if [[ "$openwisp_version" == "edge" || "$openwisp_version" == "latest" ]]; then
```
pandafy: If we do this, then there will be no different in edge and latest tags.

Makefile
```diff
@@ -1,7 +1,10 @@
 # Find documentation in README.md under
 # the heading "Makefile Options".
 
-OPENWISP_VERSION = 25.10.0
+include .env
+export
```
pandafy: Why do we need to export the environment variables? The docker compose should automatically read the OPENWISP_VERSION from the .env file.

Makefile
```diff
@@ -19,8 +22,9 @@ pull:
 	for image in 'openwisp-base' 'openwisp-nfs' 'openwisp-api' 'openwisp-dashboard' \
 				 'openwisp-freeradius' 'openwisp-nginx' 'openwisp-openvpn' 'openwisp-postfix' \
 				 'openwisp-websocket' ; do \
-		docker pull --quiet $(USER)/$${image}:$(TAG); \
-		docker tag  $(USER)/$${image}:$(TAG) openwisp/$${image}:latest; \
+				version=$${OPENWISP_VERSION:-edge}; \
+				docker pull --quiet $(USER)/$${image}:$${version}; \
+				docker tag  $(USER)/$${image}:$${version} openwisp/$${image}:$${version}; \
```
pandafy: Indentation is off!

docker-compose.yml
```diff
@@ -12,7 +12,7 @@ x-celery-depends-on: &celery-depends-on
 
 services:
   dashboard:
-    image: openwisp/openwisp-dashboard:latest
+    image: openwisp/openwisp-dashboard:${OPENWISP_VERSION:-edge}
```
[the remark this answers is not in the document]
nemesifier: @pandafy ok, I think we need to open another issue for that.

.env
```diff
@@ -4,6 +4,9 @@
 # Essential
 DASHBOARD_DOMAIN=dashboard.openwisp.org
 API_DOMAIN=api.openwisp.org
+# Image tag pinning
+IMAGE_OWNER=openwisp
+OPENWISP_VERSION=edge
```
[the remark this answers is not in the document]
atif09: the `edge` default is intentional and consistent with `TAG = edge` which is already defined in this file. this change does not produce any new behavior

users deploying to production are expected to override `OPENWISP_VERSION` in their own .env to a specific release tag

Makefile
```diff
@@ -41,11 +45,13 @@ base-build:
 	             $$BUILD_ARGS; \
 	docker build --tag openwisp/openwisp-base:latest \
 	             --file ./images/openwisp_base/Dockerfile ./images/ \
-	             $$BUILD_ARGS
+	             $$BUILD_ARGS; \
+	docker tag openwisp/openwisp-base:latest $(IMAGE_OWNER)/openwisp-base:$(OPENWISP_VERSION)
```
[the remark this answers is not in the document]
atif09: without this `docker tag` ,  the `$(IMAGE_OWNER)/openwisp-base:$(OPENWISP_VERSION)` never exists and `publish` fails

PR #603, opened 2026-05-19
[fix] Declared single source of truth for VERSION file #570

### LINKED ISSUE #570: [bug] Duplication of __openwisp_version__ leads to stale info
**Describe the bug**
The version number `__openwisp_version__` defined in `images/common/openwisp/__init__.py` is hardcoded and never updated during the release process, causing it to become stale.

Currently, it shows version `25.10.0` while the latest release is `25.10.2` (I am going to fix this shortly).

This creates version duplication across the codebase.

**Steps To Reproduce**
1. Check the current version in `images/common/openwisp/__init__.py`
2. Compare it with the latest GitHub release tag
3. Observe that the version in `__init__.py` is outdated (e.g., `25.10.0` vs `25.10.2`)

**Expected behavior**

I am not sure what's the best way to deal with this right now and I am open for suggestions.

We either need to automatically update this as part of the release process, or we need to  make this dynamic and get the right version from the information we already  have.

CodingWithSaksham:
The hardcoded `__openwisp_version__` in `openwisp/__init__.py` was never updated during the release process, causing it to report an outdated version (e.g., 25.10.0 while the actual image was 25.10.2).

This change makes the version dynamic by reading it from the OPENWISP_VERSION environment variable, which is set at Docker build time and baked into the image. This ensures `__openwisp_version__` always reflects the actual image version without manual updates.

## Checklist

Closes #570 .

## Description of Changes

- `images/openwisp_base/Dockerfile`: Create a `/opt/openwisp/OPENWISP_VERSION` file that is immutable to any writes (after final build) and write build version into this 
- `Makefile`: Change docker target flags to correctly match build stage in `images/openwisp_base/Dockerfile` and add a build arg in the last docker build to pass `OPENWISP_VERSION` into the image.
- `images/common/openwisp/__init__.py`: Replace hardcoded "25.10.0" with a `_read_version()` function that reads `/opt/openwisp/OPENWISP_VESRION` to read the version dynamically.

## Screenshot
N/A


CodingWithSaksham:
I am not sure if unit tests should be added for this kind of bug. If we add unit tests we will have to update the version in the unit tests as well which defeats the point of this PR.

images/openwisp_base/Dockerfile
```diff
@@ -49,130 +49,134 @@ RUN pip install --no-cache-dir --user --upgrade ${OPENWISP_RADIUS_SOURCE}
...
+  EMAIL_HOST_PASSWORD="" \
+  EMAIL_HOST_TLS=False \
+  EMAIL_TIMEOUT=10 \
+  EMAIL_DJANGO_DEFAULT=example@example.org \
+  DJANGO_LOG_LEVEL=ERROR \
+  DJANGO_LANGUAGE_CODE=en-gb \
+  OPENWISP_RADIUS_FREERADIUS_ALLOWED_HOSTS=172.18.0.0/16 \
+  DJANGO_X509_DEFAULT_CERT_VALIDITY=1825 \
+  DJANGO_X509_DEFAULT_CA_VALIDITY=3650 \
+  DJANGO_SECRET_KEY=DEFAULT_BAD_KEY \
```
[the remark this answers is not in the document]
CodingWithSaksham: The value is not hardcoded, its a placeholder. The user will change it according to what they want.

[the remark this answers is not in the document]
CodingWithSaksham: That is true, but this change was there even before my PR, and it has stayed that way. Also changing `DJANGO_SECRET_KEY` might cause unnecessary trouble in CI.

nemesifier (changes requested):
This PR contains changes that are totally unrelated to the fix. Please clean it up and don't send PRs for review before having double checked and cleaned your work.

CodingWithSaksham:
> This PR contains changes that are totally unrelated to the fix. Please clean it up and don't send PRs for review before having double checked and cleaned your work.

I am really sorry about the unrelated changes. I should have cleaned up the PR properly before sending it for review, and I will be more careful about that going forward.

In this case, the unrelated changes were caused by the `qa-format` formatting the Dockerfile while I was making the actual fix.  So I ended up leaving those formatting changes in the PR.

I will clean up the branch and update the PR with only the intended changes.

images/openwisp_base/Dockerfile
```diff
@@ -86,6 +86,10 @@ RUN pip install --no-cache-dir --user --upgrade -r /tmp/openwisp-deploy-requirem
     pip install --no-cache-dir --upgrade urllib3~=2.2.1
 
 FROM system
+ENV OPENWISP_VERSION="${OPENWISP_VERSION}"
+
+ARG OPENWISP_VERSION=edge
```
nemesifier: This `ARG` is good, but `make base-build` does not pass `--build-arg OPENWISP_VERSION=$(OPENWISP_VERSION)` when building `openwisp-base`. So `make release` can still build `openwisp-base:<release>` with `OPENWISP_VERSION=edge` baked in. The bug survives with a moustache. Please pass the build arg from the Makefile build path too.

images/openwisp_base/Dockerfile
```diff
@@ -86,6 +86,10 @@ RUN pip install --no-cache-dir --user --upgrade -r /tmp/openwisp-deploy-requirem
     pip install --no-cache-dir --upgrade urllib3~=2.2.1
 
 FROM system
+ENV OPENWISP_VERSION="${OPENWISP_VERSION}"
```
nemesifier: This line expands `OPENWISP_VERSION` before the final-stage `ARG` is declared, then line 92 overwrites it immediately. Please drop this dead `ENV` and keep only `ARG OPENWISP_VERSION=edge` followed by `ENV OPENWISP_VERSION="${OPENWISP_VERSION}"`.

images/common/openwisp/__init__.py
```diff
@@ -1,7 +1,9 @@
 from __future__ import absolute_import, unicode_literals
 
+import os
+
 from .celery import app as celery_app
 
 __all__ = ["celery_app"]
-__openwisp_version__ = "25.10.0"
+__openwisp_version__ = os.environ.get("OPENWISP_VERSION", "dev")
```
nemesifier: Reading from the runtime env means this reports whatever the deployment says, not necessarily the immutable image version. A stale `.env`, manual `docker run -e`, or skipped pull can make support output lie. If this value is meant to identify the image, better bake it during image build and read that baked value from Python. If we keep env-based reporting, please at least make sure the image sets it correctly at build time.

nemesifier: `dev` is a misleading production fallback. It hides a missing version as if this were intentionally a dev build. I would use `unknown`, or fail loudly if the container is expected to always define this.

nemesifier (changes requested):
The approach is OK, but not complete yet: the base image build path still does not pass the version into the Dockerfile, so release images can still bake the default value. Left the important bits inline.

CodingWithSaksham:
@coderabbitai review

images/common/openwisp/__init__.py
```diff
@@ -1,7 +1,16 @@
...
+
 from .celery import app as celery_app
 
 __all__ = ["celery_app"]
-__openwisp_version__ = "25.10.0"
+__openwisp_version__ = os.environ.get("OPENWISP_VERSION")
 __openwisp_installation_method__ = "docker-openwisp"
+
+__openwisp_version__ = os.environ.get("OPENWISP_VERSION")
+if __openwisp_version__ is None or __openwisp_version__ == "unknown":
```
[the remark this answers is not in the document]
CodingWithSaksham: Already fixed in new commit

images/common/openwisp/__init__.py
```diff
@@ -1,7 +1,15 @@
 from __future__ import absolute_import, unicode_literals
 
+import os
+
 from .celery import app as celery_app
 
 __all__ = ["celery_app"]
-__openwisp_version__ = "25.10.0"
+__openwisp_version__ = os.environ.get("OPENWISP_VERSION")
```
nemesifier: This still reads the mutable runtime env, so compose can override the value baked into the image. I tested with `make compose-build OPENWISP_VERSION=26.06a` and then `OPENWISP_VERSION=26.06a docker compose run --rm --no-deps --entrypoint python dashboard -c 'import openwisp, os; print(os.environ.get("OPENWISP_VERSION")); print(openwisp.__openwisp_version__)'`. Expected: `26.06a` / `26.06a`. Actual: `edge` / `edge`, because `.env` contains `OPENWISP_VERSION=edge` and `env_file` wins at runtime. Please bake the build arg into an immutable file/module and read that instead.

images/common/openwisp/__init__.py
```diff
@@ -1,7 +1,15 @@
...
 __openwisp_installation_method__ = "docker-openwisp"
+
+if __openwisp_version__ is None or __openwisp_version__ == "unknown":
```
nemesifier: Raising from `openwisp/__init__.py` makes every package import fail if the env is missing or set unexpectedly. If the version is baked into a file/module at build time, this import-time guard should not be needed; if we keep a guard, I would put it in a narrower build/test verification path.

Makefile
```diff
@@ -48,8 +48,10 @@ base-build:
 	             --target PYTHON ./images/ \
```
nemesifier: While testing `make compose-build OPENWISP_VERSION=26.06a`, this stage printed `ERROR: failed to solve: target stage "PYTHON" could not be found` but the recipe continued and built the final images. Expected behavior: fail immediately on any failed Docker build step, or use a valid target and print no error. This may be pre-existing, but this PR now relies on this build path for version correctness.

nemesifier (changes requested):
Thanks for addressing the first round, this is definitely moving in the right direction. I tested the build/runtime path locally and there is still one important correctness issue: the version is baked correctly, but compose can override it at runtime via `.env`, so `__openwisp_version__` can still report the wrong value. I think the clean fix is to bake the version into an immutable file/module during the image build and read that from Python.

Makefile
CodingWithSaksham: It appears
```Dockerfile
	docker build --tag openwisp/openwisp-base:intermedia-system \
	             --file ./images/openwisp_base/Dockerfile \
	             --target SYSTEM ./images/; \
```

was pointing to the wrong path as well, I will be adding a fix for both of these in this PR only.
I wonder how this issue went under the radar for so long :thinking:

images/common/openwisp/__init__.py
CodingWithSaksham: Do we also want to set `OPENWISP_VERSION` as an environment variable in the Dockerfile for the user, or is setting only `__openwisp_version__` to the correct version sufficient?

CodingWithSaksham:
I am not sure why the CI keeps failing, whenever its re-run the CI passes.

nemesifier (changes requested):

nemesifier (commented):
I think the approach we had before was not bad, the problem is that it could be overridden with an env var which is wrong. 

Any way, if you can't think of a better way to do this give me some more time to try this and come to a conclusion.

nemesifier:
@coderabbitai full review

CodingWithSaksham:
> I think the approach we had before was not bad, the problem is that it could be overridden with an env var which is wrong.
> 
> Any way, if you can't think of a better way to do this give me some more time to try this and come to a conclusion.

Lets discuss the approach for this first, for now I have converted the PR to a draft PR, once we agree on a final approach I'll make it ready for review

Multiple approaches do come to my mind

1. To manually substituting `__openwisp_version__` at runtime when we build the docker image. This can be done by substituting the previous value with the new value via regex substitution (maybe through `build.py`).
2. At build time, generate a `images/common/openwisp/_version.py` file containing just `__version__ = "26.01a"`. `__init__.py` imports from it.
3. Package the app properly with setup.cfg or pyproject.toml and read version via `importlib.metadata.version("openwisp")`. 

And honestly? The 2nd option seems more promising to me as I have seen libraries like numpy and pandas do the same, it is similar to our current approach but less flaky.

CodingWithSaksham:
@coderabbitai full review

images/openwisp_base/Dockerfile
```diff
@@ -6,20 +6,20 @@ FROM python:3.14-slim-trixie AS system
 # 3. openwisp-monitoring: fping gdal-bin
 # hadolint ignore=DL3008
 RUN apt-get update && \
-    apt-get install --yes --no-install-recommends \
```
nemesifier, line 9: avoid unrelated changes

nemesifier (changes requested):

images/common/openwisp/__init__.py
```diff
@@ -1,7 +1,10 @@
 from __future__ import absolute_import, unicode_literals
 
+# This is autogenerated by Docker when running
+# make base-build or make runtests
+from ._version import __version__ as __openwisp_version__
```
nemesifier: After thinking this through, I do not think the generated ignored `_version.py` approach is acceptable. The version must be part of the repository, not an untracked build artifact. Otherwise a clean checkout cannot import `openwisp`, code review cannot see the source of truth, and the release process can silently build whatever happened to be generated locally.

Please change the approach to a tracked file, for example:

```python
# images/common/openwisp/_version.py
__version__ = "25.10.0"
```

Then `__init__.py` can import it directly. Add a Makefile target such as `make update-version VERSION=25.10.3` which updates both `RELEASE_VERSION` in `Makefile` and `images/common/openwisp/_version.py` in the same command. Document that target in the developer or release docs. The Docker build should only copy the tracked file into the image, not generate an ignored module at build time.

build.py
```diff
@@ -31,17 +31,71 @@ def get_secret_key(allow_special_chars=True):
     return keygen
 
 
+def update_makefile_version(new_version: str):
```
nemesifier: This helper points in the right direction, but the implementation is incomplete. Updating only `RELEASE_VERSION` still leaves the Python version as a separate value, which is exactly the duplication this issue is about. It is also not wired to any Makefile target, so it is dead code in the current PR.

Please make this the single version update path instead: a Makefile target like `make update-version VERSION=25.10.3` should call this helper, and the helper should update both `RELEASE_VERSION` in `Makefile` and the tracked `images/common/openwisp/_version.py`. It should fail if either file cannot be updated. Then document this target in the release/developer docs so maintainers know how to bump the version.

nemesifier (commented):
I do not think the current generated ignored `_version.py` approach is the right solution. The version should be stored in a tracked file in the repo, updated by a Makefile target, and documented in the release/developer docs. Otherwise a clean checkout can fail to import `openwisp`, the source of truth is not reviewable, and the release output depends on a local generated artifact.

.gitignore
```diff
@@ -82,3 +82,5 @@ local/
 customization/*
 !customization/configuration/django/__init__.py
 !customization/configuration/django/custom_django_settings.example.py
+
+_version.py
```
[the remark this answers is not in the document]
nemesifier: I want to revise this: please remove this ignore rule entirely. `_version.py` should not be generated and ignored. It should be a tracked repository file containing the current OpenWISP version, updated by a Makefile target during releases.

build.py
```diff
@@ -31,17 +31,72 @@ def get_secret_key(allow_special_chars=True):
...
+
+def generate_version_module(version: str):
+    # Generate images/common/openwisp/_version.py with the given version.
+
+    version_file = "images/common/openwisp/_version.py"
+    content = (
+        "# This file is auto-generated at Docker image build time.\n"
+        "# Do not edit manually.\n"
+        f'__version__ = "{version}"\n'
+    )
```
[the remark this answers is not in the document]
nemesifier: This becomes less important if we stop generating Python source during `base-build`, which I think we should. The better fix is to keep `images/common/openwisp/_version.py` tracked in the repo and update it through `make update-version VERSION=...`. If the helper writes that file, it should still use a safe Python literal such as `{version!r}` and fail if the update did not happen.

.gitignore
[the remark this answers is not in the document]
nemesifier: This is stale after my updated comment above. Please do not scope the ignore rule. Remove it entirely, because `images/common/openwisp/_version.py` should be tracked in the repository.

build.py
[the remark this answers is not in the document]
nemesifier: This is stale after my updated comment above. Please do not keep generating this file during `base-build`. The requested approach is a tracked `images/common/openwisp/_version.py`, updated by a Makefile target and documented in the release/developer docs.

nemesifier:
@coderabbitai full review

nemesifier (approved):
@CodingWithSaksham I pushed some changes which I tested manually in my local env and seem to be working fine, please double check on your end.

nemesifier:
@coderabbitai full review

tests/runtests.py
```diff
@@ -528,5 +530,54 @@ def test_containers_down(self):
...
+            makefile.write_text((repository_root / "Makefile").read_text())
+            (tmpdir / ".env").write_text("")
+            (tmpdir / "build.py").write_text((repository_root / "build.py").read_text())
+            result = subprocess.run(
+                ["make", "bump", "VERSION=26.01.0"],
+                cwd=tmpdir,
+                check=False,
+                capture_output=True,
+                text=True,
+            )
```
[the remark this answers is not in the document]
nemesifier: No thanks. Focus only on real issues.

nemesifier:
@coderabbitai resolve

PR #610, opened 2026-05-20
[change] Avoid unnecessary CRL restarts when revokelist content is unchanged #589

### LINKED ISSUE #589: [change] OpenVPN revokelist.sh script seems to needlessly restart OpenVPN
I see a number of problems:

1. The revokelist.sh implementation si very naive, it always downloads the CRL and restarts OpenVPN every night.
We should instead check if either the CRL or VPN config has been changed and only then restart the VPN, we can then change the frequency to so it can run more often (eg: 5 minutes).

2. The crontab in the openvpn container is created by scripts and not clearly visible in the image.
I wonder if we can keep the crontab in the image dir to ease maintenance?

3. Similar to 2, the bash functions which are used in the openvpn image are stored in the common directory, which doesn't make any sense to me, as this logic is specific to the openvpn image and should be self contained in it to ease maintenance.

4f4d:
## Checklist

- [x] I have read the OpenWISP Contributing Guidelines.

Closes #589

## Description of Changes

- moved the OpenVPN-specific helper functions from `images/common/utils.sh` to `images/openwisp_openvpn/openvpn_utils.sh`
- moved the base OpenVPN cron entries into `images/openwisp_openvpn/openvpn.crontab` and load them from `init_command.sh`
- updated `revokelist.sh` to download the CRL into a temporary file and restart OpenVPN only when the CRL actually changes
- changed the CRL refresh frequency from once per day to every 5 minutes now that unchanged CRLs no longer trigger restarts
- validated `TOPOLOGY_UPDATE_INTERVAL` before appending the optional topology cron entry
- hardened Redis reads and OpenVPN checksum/config downloads to fail fast on invalid or incomplete data
- added regression coverage for the CRL restart behavior and made the affected test paths retry-safe in CI

## Local Validation

- `./run-qa-checks` completed successfully
- `make compose-build` completed successfully, including the patched OpenVPN image
- validated the OpenVPN-specific runtime behavior in a container:
  - the image-local `openvpn.crontab` is loaded correctly
  - the first CRL download creates `revoked.crl` and triggers the restart path
  - an unchanged CRL does not trigger a restart
  - a changed CRL replaces `revoked.crl` and triggers the restart path
  - a failed CRL download preserves the existing `revoked.crl` and does not trigger a restart


4f4d:
@coderabbitai review

4f4d:
@coderabbitai review

images/openwisp_openvpn/openvpn_utils.sh
```diff
@@ -0,0 +1,137 @@
...
+	fi
+}
+
+function openvpn_config_checksum {
+	local remote_checksum
+	local local_checksum
+
+	remote_checksum=$(curl --silent --show-error --fail --insecure \
+		${API_INTERNAL}/controller/vpn/checksum/$UUID/?key=$KEY) || return 1
+	if [ -z "$remote_checksum" ] || [ ! -s checksum ]; then
```
nemesifier: This treats a missing or empty local `checksum` the same as a failed remote checksum fetch. I tested the missing-local-checksum case by stubbing the remote checksum to return `remote-checksum` and leaving local `checksum` absent. Expected behavior: download the config, update `checksum`, and restart OpenVPN. Actual behavior: `openvpn.sh` exits with `Failed to fetch or validate OpenVPN checksum`, does not download anything, and does not restart.

We must still fail if the remote checksum cannot be downloaded or is empty, but a missing local checksum should mean local state is stale/corrupt and should trigger a safe config re-download.

images/openwisp_openvpn/openvpn_utils.sh
```diff
@@ -0,0 +1,137 @@
...
+	curl --silent --show-error --fail --retry 10 --retry-delay 5 --retry-max-time 300 \
+		--insecure --output "$tmp_tar" \
+		${API_INTERNAL}/controller/vpn/download-config/$UUID/?key=$KEY || return 1
+	test -s "$tmp_tar" || return 1
+
+	curl --silent --show-error --fail --insecure --output "$tmp_checksum" \
+		${API_INTERNAL}/controller/vpn/checksum/$UUID/?key=$KEY || return 1
+	test -s "$tmp_checksum" || return 1
+
+	tar xzf "$tmp_tar" || return 1
```
nemesifier: This extracts the downloaded tarball directly into `/` before verifying the downloaded config against the checksum returned by the API. We need an integrity check before applying anything.

Expected behavior:

```text
- download config and checksum into temporary paths
- verify the downloaded config checksum matches the expected checksum
- only then replace the live files and restart OpenVPN
```

If config download fails, checksum download fails, checksum is empty, or the downloaded config checksum does not match the API checksum, we must log an error, keep the existing config/checksum untouched, and not restart OpenVPN. Otherwise a partial/corrupted/mismatched config can be applied and later picked up by the next service/container restart.

images/openwisp_openvpn/openvpn.sh
```diff
@@ -3,12 +3,19 @@
...
 openvpn_config
-openvpn_config_checksum
+if ! openvpn_config_checksum; then
+	echo "Failed to fetch or validate OpenVPN checksum" >&2
+	exit 1
+fi
 
 if [ "${OFILE}" != "${NFILE}" ]; then
-	openvpn_config_download
+	if ! openvpn_config_download; then
```
nemesifier, line 16: Please add coverage for the config integrity path too. The CRL regression test is useful, but we also need to simulate a downloaded config/checksum mismatch and assert that the old config remains in place, `checksum` remains unchanged, `supervisorctl restart openvpn` is not called, and an error is logged. I am sorry this was not specified in high detail before.

nemesifier (changes requested):
Thanks for the work on this, the CRL part is going in the right direction. I built the OpenVPN image locally, verified `/openvpn.crontab` loads correctly, and exercised `revokelist.sh` with controlled CRL contents. Initial/changed CRL restarts OpenVPN, unchanged CRL does not restart, and failed CRL download keeps the existing `revoked.crl`, which is what we want.

I am requesting changes on the VPN config update path: we need to be stricter about applying downloaded configs and also preserve the useful self-recovery behavior when local state is missing/corrupt.

images/openwisp_openvpn/openvpn_utils.sh
4f4d: Addressed in f543df0

images/openwisp_openvpn/openvpn_utils.sh
4f4d: Addressed in f543df0

images/openwisp_openvpn/openvpn.sh
4f4d: Added config-path regression coverage in `tests/runtests.py`.

The new test covers:
- checksum mismatch preserving the existing config/checksum and avoiding restart
- missing local checksum triggering safe re-download and restart

The updated branch is green.

nemesifier:
@coderabbitai full review

4f4d:
@coderabbitai full review

4f4d:
@coderabbitai full review

images/openwisp_openvpn/openvpn_utils.sh
```diff
@@ -0,0 +1,180 @@
...
+	# Fectch UUID and Key of the default VPN only if they
+	# are not already set. The user may override the UUID and Key
+	# by setting them in the environment variables to use deploy
+	# a different VPN server.
+	if [ -z "$UUID" ]; then
+		export UUID=$(get_redis_value "openwisp_default_vpn_uuid")
+		export KEY=$(get_redis_value "openwisp_default_vpn_key")
+		export CA_UUID=$(get_redis_value "openwisp_default_vpn_ca_uuid")
+	fi
+}
```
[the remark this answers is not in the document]
nemesifier: This is a good suggestion. The current values are expected to be UUIDs/keys, so this is not the main blocker, but quoting the substitutions is the cheap safe form and avoids word splitting if a value is ever malformed. I would split assignment/export as suggested.

images/openwisp_openvpn/openvpn_utils.sh
```diff
@@ -0,0 +1,180 @@
...
+	curl --silent --show-error --fail --retry 10 --retry-delay 5 --retry-max-time 300 \
+		--insecure --output "$output_path" \
+		${DASHBOARD_INTERNAL}/admin/pki/ca/x509/ca/${CA_UUID}.crl
+}
+
+function crl_download {
+	curl --silent --show-error --fail --retry 10 --retry-delay 5 --retry-max-time 300 \
+		--insecure --output revoked.crl \
+		${DASHBOARD_INTERNAL}/admin/pki/ca/x509/ca/${CA_UUID}.crl
+}
```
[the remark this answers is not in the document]
nemesifier: This is a good suggestion too, especially if we add validation to the CRL download path. Keeping two curl implementations makes it easy to harden the refresh path and accidentally leave startup with different behavior. I would make crl_download a small wrapper around crl_download_to.

images/openwisp_openvpn/openvpn_utils.sh
```diff
@@ -0,0 +1,180 @@
...
+
+function crl_download_if_changed {
+	local tmp_crl
+	tmp_crl=$(mktemp /tmp/revoked.crl.XXXXXX) || return 2
+	trap 'rm -f "$tmp_crl"' EXIT HUP INT TERM
+
+	if ! crl_download_to "$tmp_crl"; then
```
nemesifier, line 157: Can we treat an empty CRL as a failed download here? `curl --fail` only catches HTTP errors. If the dashboard returns 200 with an empty body, this function will promote that empty file at line 150 and restart OpenVPN, which can temporarily disable revocation checks. I would add a `test -s "$tmp_crl"` check before comparing/promoting, return 2 on empty, keep the existing `revoked.crl`, and add a small case to the regression test.

nemesifier (changes requested):
I think the main approach solves #589: comparing the downloaded CRL before restarting is the right direction, and moving the OpenVPN cron/helpers into the image makes this easier to maintain. I found one CRL validation gap that I would fix before merging.

nemesifier:
@coderabbitai full review

nemesifier (commented):
Looks good, but I haven't tested it yet, I need to do one manual run to ensure it works as expected and then we can merge.

nemesifier (changes requested):
Are you sure this works with the real CRL endpoint?

I tested this branch in a rebuilt live `openvpn` container. The synthetic cases behave as expected: first CRL download restarts OpenVPN, unchanged content does not restart, changed content restarts, and failed or empty downloads preserve the existing CRL without restarting.

However, when I tested the real CRL endpoint, I got a restart on consecutive runs even though I was expecting the second run to detect no meaningful CRL change.

I downloaded the CRL twice a few seconds apart and got different hashes:

```text
one.crl sha256: 5ddfe4d356f89490d064735de385ea6182ebfdd5767e3cad387396ee0b8bd002
two.crl sha256: 9b61fc47d725ba1d59ada6a086910f6b9a02b770fbe8231ade122db9dd94cf4e
CMP=different
```

The difference seems to be caused by CRL metadata being regenerated on each request:

```text
ONE:
lastUpdate=Jul 18 19:56:26 2026 GMT
nextUpdate=Jul 19 19:56:26 2026 GMT

TWO:
lastUpdate=Jul 18 19:56:28 2026 GMT
nextUpdate=Jul 19 19:56:28 2026 GMT
```

The revoked certificate section looked unchanged in my test:

```text
CRL_one_REVOKED:
CRL_two_REVOKED:
```

So with the current `cmp -s "$tmp_crl" revoked.crl` logic, are we still going to restart OpenVPN every time `/revokelist.sh` runs, now every 5 minutes instead of once per day?

Could you please verify this against a live stack and adjust the change detection so timestamp-only CRL changes do not trigger an OpenVPN restart?

nemesifier:
Superseded by https://github.com/openwisp/docker-openwisp/pull/674.

Issue #582, opened 2026-03-30
[bug] Investigate random hangs and crashes in Selenium tests with websockets on GitHub Actions


pandafy:
**Describe the bug**
The test suite intermittently fails on GitHub Actions, specifically with tests involving websockets. Upon investigation, when running Selenium-based tests, they randomly hang and then crash with a `ReadTimeoutError`. This typically occurs when the Selenium Python client sends a command to ChromeDriver but never receives a response back. It appears the ChromeDriver process hangs without feedback, which leads to test failures.

This has impacted the ability to reliably run end-to-end or integration tests in CI. The issue was observed and discussed in [PR #581](https://github.com/openwisp/docker-openwisp/pull/581), where the failing tests were removed temporarily to allow the CI to pass.

**Steps To Reproduce**

1. Run the Selenium test suite in GitHub Actions, especially tests relying on websockets
2. Observe intermittent test hangs and failures
3. Check the logs for `ReadTimeoutError` from Selenium
4. Note that the process never gets a response from ChromeDriver

**Expected behavior**
After adding back the tests removed in #581, all Selenium-driven tests (including those that use websockets) should complete without any unexplained timeout or hang. ChromeDriver should remain responsive and receive/respond to all commands.

- OS: Ubuntu (GitHub Actions latest runner)
- Docker version: Please specify from CI logs
- Browser and Browser Version (if applicable): Chrome (specify from CI logs)

_Reference: The failing tests were removed in [PR #581](https://github.com/openwisp/docker-openwisp/pull/581) due to this issue. Please investigate and propose a robust solution to enable reliable testing in CI._

Issue #580, opened 2026-03-27
[chores] Make running tests atomic to prevent database corruption on failure


pandafy:
### Problem
Currently, running the test suite is not atomic. If any test fails during execution, the database can be left in a broken or inconsistent state, which affects further test runs and developer productivity.

### Expected behavior
All tests should  be isolated in a way that ensures the database remains consistent, regardless of test outcomes. Failed test runs should not leave the database in a broken state.

### Suggested solution
- Ensure each test either uses a transactional database setup or is properly cleaned up after execution.
- Consider using fixtures or a setup/teardown process that restores the database state after every test or test suite.
- Investigate test runner options or custom scripts to enforce atomicity for database operations during tests.

### Additional context
Making the test suite atomic will make development safer and more reliable for contributors, avoiding hidden issues due to database state leaks between test runs.

CodingWithSaksham:
One of the few things that I have noticed is
1) Running `make runtests` twice in a row will lead to the same issue as described in by @nemesifier in #630 and is fixed by #621 
2) Running selenium tests sometimes causes one of the tests (I am not sure which one) to be stuck, it requires manual refreshing of the page for the test to succeed. If this issue is solved, it will drastically improve the CI as well, because most of the selenium failures are due to this only. 
3)  Running the tests locally on a Linux machine will cause the tests to fail due to missing `/etc/hosts`. Will have to check if this is just for Fedora (my os) or every Linux machine #614 
4) `test_add_superuser` fails if chrome/firefox window is smaller than 500x500 pixels. 

I will keep adding more to this list if I find new ones, but the major one is the 2nd one.

nemesifier:
@CodingWithSaksham thanks for your analysis, the best thing is to start with some low hanging fruits.

nemesifier:
The concrete stale-database scenario described here has been addressed by #621.

The fix cleans up test-created users before each functional test run and ensures CI tears down the stack once the retry loop finishes, including after a failed attempt. This prevents a transient Selenium failure from causing subsequent retries or later runs to fail due to leftover users.

I also ran the functional test suite twice consecutively on the current `master` branch. Both runs passed, including the user-creation and registration tests that previously exposed stale database state.

Since the reported behavior is no longer reproducible and the known cause has been fixed, this issue can be closed.

PR #620, opened 2026-05-29
[fix] Dynamically inject __openwisp_version__ into immutable file at build time

### LINKED ISSUE #570: [bug] Duplication of __openwisp_version__ leads to stale info

akshatj0630:
## Checklist

- [x] I have read the OpenWISP Contributing Guidelines.

Closes #570.

## Description of Changes

This PR resolves the version staleness issue by establishing a robust, immutable way to inject `__openwisp_version__` into the Docker images without triggering syntax injection risks or UI test timeouts.

1. **Dockerfile:** Used `ARG OPENWISP_VERSION` and `printf` to write the version directly into a plain text `/opt/openwisp/openwisp/.version-info` file during the final build step.
2. **Makefile:** Fixed the broken `--target SYSTEM` and `--target PYTHON` flags (changed to `system` and `openwisp_python`) so the Docker cache correctly picks up the stages, and explicitly passed `--build-arg OPENWISP_VERSION`.
3. **__init__.py:** Replaced the hardcoded version with a safe `try/except FileNotFoundError` block that reads `.version-info`. This strictly eliminates `.env` runtime override bugs, avoids Python string-escaping vulnerabilities, and gracefully degrades to `"unknown"` outside of Docker to prevent Selenium tests from throwing `ImportError` timeouts in CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,13 +42,14 @@ base-build:
 	done; \
 	docker build --tag openwisp/openwisp-base:intermedia-system \
 	             --file ./images/openwisp_base/Dockerfile \
-	             --target SYSTEM ./images/; \
+	             --target system ./images/; \
 	docker build --tag openwisp/openwisp-base:intermedia-python \
 	             --file ./images/openwisp_base/Dockerfile \
-	             --target PYTHON ./images/ \
+	             --target openwisp_python ./images/ \
 	             $$BUILD_ARGS; \
 	docker build --tag $(IMAGE_OWNER)/openwisp-base:$(OPENWISP_VERSION) \
 	             --file ./images/openwisp_base/Dockerfile ./images/ \
+	             --build-arg OPENWISP_VERSION=$(OPENWISP_VERSION) \
 	             $$BUILD_ARGS
 
 nfs-build:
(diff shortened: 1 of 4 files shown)


akshatj0630:
Note: The earlier CI failures (Attempts 2-5 on the previous commit) were caused by the known transient Selenium/websocket hangs during teardown, which poisoned the retry attempts with duplicate users in the database (as tracked in issues #582 and #580). I triggered a fresh rebuild and it passed perfectly.

(Expanding the exception handler at tests/runtests.py:189 to catch Exception rather than just NoSuchElementException would likely prevent unexpected timeouts like ReadTimeoutError from breaking teardown and poisoning the subsequent retries.)

nemesifier:
This issue is address to someone else, closing as duplicate.

PR #674, opened 2026-08-25
[change] Made CRL reload smarter to avoid daily openvpn restart

### LINKED ISSUE #589: [change] OpenVPN revokelist.sh script seems to needlessly restart OpenVPN

nemesifier:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html) and [OpenWISP Anti AI Spam Policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-ai-spam-policy).

Closes #589.

## Description of Changes

- move OpenVPN helpers and base cron entries into the image directory
- refresh the CRL every five minutes without restarting for timestamp-only changes
- preserve the existing CRL when downloads fail, are empty, or cannot be parsed
- restart OpenVPN when revoked certificate entries change
- add container-level regression coverage for CRL refresh behavior

## Screenshot

N/A


docs/user/settings.rst
```diff
@@ -1037,6 +1037,14 @@ OpenVPN
 - **Valid Values:** STRING.
 - **Default:** ``default-management-vpn``.
 
+Revocation List Refresh
```
nemesifier: @coderabbitai I think this is out of context for this settings page.
This is how it's supposed to work and users shouldn't be concerned with these details. This kind of detail can be kept in code comments.

nemesifier:
@coderabbitai resolve

PR #675, opened 2026-08-26
[fix] Fixed small flaws in auto-install bash logic #588

### LINKED ISSUE #588: [bug] auto-install.sh: pre-existing issues in get_env, set_env and USE_OPENWISP_RADIUS logic
Found during review of #562 fix. These are pre-existing issues unrelated to that change.

## 1. `get_env` uses prefix matching

```bash
get_env() { grep "^$1" "$2" | cut -d'=' -f 2-50; }
```

`grep "^$1"` matches any variable whose name **starts with** the given string. For example, `get_env "DB_PASS"` would also match `DB_PASSWORD` if that variable existed. Should be:

```bash
get_env() { grep "^$1=" "$2" | cut -d'=' -f 2-50; }
```

## 2. `set_env` uses `eval` on unsanitized values

```bash
eval $(echo "awk -i inplace 'NR=="${line_number}" {\$0=\"${1}=${2}\"}1' $INSTALL_PATH/.env")
```

If `$1` (variable name) or `$2` (value) contains shell metacharacters, this can execute arbitrary code. Values ultimately come from user input or the `.env` file. Should use a safer `awk` invocation that passes values via `-v` instead of interpolating them into the command string.

## 3. `USE_OPENWISP_RADIUS` logic appears inverted

```bash
if [[ -z "$USE_OPENWISP_RADIUS" ]]; then
    set_env "USE_OPENWISP_RADIUS" "Yes"
else
    set_env "USE_OPENWISP_RADIUS" "No"
fi
```

When `$USE_OPENWISP_RADIUS` is unset/empty, Radius is enabled (`Yes`). When it has any value, it is disabled (`No`). This seems backwards — needs verification.

nemesifier:
## Checklist

- [x] I have read the [OpenWISP Contributing Guidelines](https://openwisp.io/docs/dev/developer/contributing.html) and [OpenWISP Anti AI Spam Policy](https://openwisp.io/docs/dev/general/code-of-conduct.html#anti-ai-spam-policy).
- [ ] I have manually tested the changes proposed in this pull request.
- N/A I have written new test cases for new code and/or updated existing tests for changes to existing code.

Closes #588.

## Description of Changes

Fixed exact environment-variable lookups, replaced shell evaluation with a literal configuration-file rewrite, and validated the Radius setting against the documented `True` and `False` values.

## Screenshot

N/A

### DIFF
--- a/deploy/auto-install.sh
+++ b/deploy/auto-install.sh
@@ -17,14 +17,27 @@ export NON='\033[0m'
 start_step() { printf '\e[1;34m%-70s\e[m' "$1" && echo "$1" &>>$LOG_FILE; }
 report_ok() { echo -e ${GRN}" done"${NON}; }
 report_error() { echo -e ${RED}" error"${NON}; }
-get_env() { grep "^$1" "$2" | cut -d'=' -f 2-50; }
+# Match the variable name exactly to avoid reading similarly named settings.
+get_env() { grep "^$1=" "$2" | cut -d'=' -f 2-50; }
+# Update values without evaluating configuration or user input as shell code.
 set_env() {
-	line=$(grep -n "^$1=" $INSTALL_PATH/.env)
-	if [ -z "$line" ]; then
-		echo "$1=$2" >>$INSTALL_PATH/.env
+	local env_file=$INSTALL_PATH/.env
+	local temp_file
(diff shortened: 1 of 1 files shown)


nemesifier:
@coderabbitai review

